PR #2, opened 2019-06-24
Dockerdevel

francescalb:
Added Dockerfile for development purposes as well as documentation in doc for this.


jesper-friis:
Seems ok

PR #13, opened 2019-10-02
Fixed #5 - homogenised call to reasoner

### LINKED ISSUE #5: Homogenise call to reasoner in emmo.Ontology.sync_reasoner()
Pellet should be the default

francescalb:
Jesper fixed choice of reasoner.


emmo/ontology.py
```diff
@@ -159,10 +159,28 @@ def get_by_label_all(self, label):
                     getattr(self, c)() for c in categories)
                 if hasattr(entity, 'label') and label in entity.label]
 
-    def sync_reasoner(self):
-        """Update current ontology by running the HermiT reasoner."""
-        with self:
-            owlready2.sync_reasoner()
+    def sync_reasoner(self, reasoner='Pellet', include_imported=False):
```
jesper-friis: Maybe HermiT should be the default

PR #19, opened 2020-01-13
1.0.0

jesper-friis:
I think it is time to merge the EMMO-python 1.0.0 into master even though we haven't released EMMO v1.0.0 yet. The reason is that master is very old and outdated. We could release it as version 1.0.0-alpha-1.


jesper-friis:
We can leave master in this miserable state. I merge now...

PR #20, opened 2020-02-28
Updated inferred ontology to latest version of EMMO from the 1.0.0 branch.

jesper-friis:


jesper-friis:
The inferred ontology is now published together with EMMO. I close this.

PR #34, opened 2020-04-02
Updated simplifed demo_vertical in compliance with EMMO-1.0.0alpha2 as of 202…

francescalb:
…0-03-31


jesper-friis:
Works fine. The plots produced by plot_ontology.py should be revised. For example make units+properties.svg and types+properties.svg not sense any more.

jesper-friis:
Great. I think we can merge to master and create a new release.

PR #41, opened 2020-08-12
Installed missing dependencies in pythonpublish deployment workflow

jesper-friis:


francescalb:
This is good.

francescalb:
This looks good

PR #42, opened 2020-08-18
Ensure that emmocheck exit with non-zero return value if a test is faing

jesper-friis:


francescalb (commented):
This works well

francescalb (approved):
Approved

PR #43, opened 2020-09-16
Update emmodoc to latest version of emmo-alpha2

jesper-friis:
PhysicsDimension is now PhysicalDimension

This update is not perfect since it breaks backward compatibility, but needed for generating documentation of upstream EMMO...


francescalb (commented):
Need version-update to .13

PR #44, opened 2020-09-18
Make emmo package pip installable in fresh env

### LINKED ISSUE #45: Not immediately installable with pip
When installing this with the `pip` package manager, it fails due to the import of itself (`emmo`).

This can be fixed by avoiding the import of itself and instead retrieve the necessary information "manually", i.e., through raw file access, scanning, and parsing.

CasperWA:
Fixes #45.

Don't import `emmo` package prior to installation, instead retrieve the version through manual retrieval from `emmo/__init__.py`.

Change version variable to the more "pythonic" `__version__`.


jesper-friis (approved):
Thanks Casper. Looks good.

PR #51, opened 2020-09-23
Added all labels in get_class_annotations in emmo/patch.py including

francescalb:
Added all annotations including altLabel in emmo/patch.py.
Gerenation of htmldocumentation proceeds correctly locally now, when testing on latest version of emmo in branch 'add-additional-quantities'. Note that I had to remake the inferred ontology locally.


jesper-friis (approved):
Looks fine

PR #56, opened 2020-10-24
Correct default IRI to inferred ontology

jesper-friis:
The "unstable" symlink is deprecated and replaced with "development". However, it make more sense that the default IRI to the inferred ontology refer to "latest-stable".


francescalb (approved):
Tested successfully for all three emmo-version (emmo, latest emmo-inferred and development emmo-inferred)

PR #57, opened 2020-10-25
Added get_preferred_label() method to classes, properties and individuals

jesper-friis:
Updated patch.py with following:
- added get_preferred_label() method to classed and properties
- added get_preflabel() method to individuals (cannot use same name as for classes)
- added get_typename() method to restrictions

### DIFF
--- a/emmo/patch.py
+++ b/emmo/patch.py
@@ -1,7 +1,9 @@
 # -*- coding: utf-8 -*-
 """This module injects some additional methods into owlready2 classes."""
+from types import MethodType
+
 import owlready2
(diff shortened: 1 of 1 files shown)


francescalb:
Works correctly. A suggestion: Maybe put in an error message for get_preflabel (if self is is a class or property) if preffered_label should be used and the same for the other way around?

francescalb:
> Works correctly. A suggestion: Maybe put in an error message for get_preflabel (if self is is a class or property) if preffered_label should be used and the same for the other way around?

Bad suggestion by me. Not possible to do. I suggest to merge and have the error message in mind for later .

PR #60, opened 2020-10-30
Added --ignore-namespace to emmocheck 

francescalb:
It is now possible to ignore the a given namespace by:

emmocheck --ignore-namespace=http://emmo.info/emmo/emmo-inferred# emmo-inferred

However, this is not the same as ignoring submodules so does not work correctly yet. When  you load an ontology, the base_iri is the same for the whole ontology...

### DIFF
--- a/emmo/emmocheck.py
+++ b/emmo/emmocheck.py
@@ -212,8 +212,6 @@ class TestFunctionalEMMOConventions(TestEMMOConventions):
         ))
         exceptions.update(self.get_config('test_namespace.exceptions', ()))
         def checker(onto,ignore_namespace):
-            print(onto)
-            print(ignore_namespace)
             print(list(itertools.chain(onto.classes(),
                                      onto.object_properties(),
                                      onto.data_properties(),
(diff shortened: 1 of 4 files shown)


emmo/emmocheck.py
```diff
@@ -281,6 +283,10 @@ def main():
     parser.add_argument(
         '--url-from-catalog', '-u', action='store_true',
         help=('Get url from catalog file'))
+    parser.add_argument(
+        '--ignore-namespace', '-in', action='append', default=[],
```
jesper-friis, line 287: Unusual with 2-letter short option. Normally -in is a shortcut for "-i -n" if "-i" doesn't take an argument or "-i n" if "-i" takes an argument.

jesper-friis (commented):
See inline comment

emmo/emmocheck.py
francescalb: thank you, this is fixed

francescalb:
Works for skipping namespace check for 'emmocheck --check-imported --ignore-namespace=emmo-inferred emmo-inferred'. However namespace check for emmo does not go into the check at all because when onto = get_ontolog('emmo').load() all of the following are empty: onto.classes(),                         onto.object_properties(), onto.data_properties(), onto.individuals(), onto.annotation_properties(). Is this correct behaviour?

jesper-friis (approved):
Great Francesca. Thanks

PR #61, opened 2020-11-01
Improve support for imported ontologies

jesper-friis:
Adding option `imported` to several methods.


francescalb (commented):
This works correctly. It fail when checking autimatically imported emmo-inferred, but this is due to an error in the ontology and not emmocheck. Approved.

francescalb (approved):

PR #63, opened 2020-11-07
Improved error messages for classes that doesn't define prefLabel

jesper-friis:
Also improved docstring for plotting module dependencies.


francescalb (approved):
Better error message is shown. It could be even more improved by including which types of labels that are included.

jesper-friis:
Test fails while installing latex dependencies due to unacceptable server. 

Not sure what to do in this case. We would like that the tests passes on master - but the reason for the failure is not due to EMMO-python.

francescalb:
Fixed

PR #64, opened 2020-11-20
Added ontoconvert tool

jesper-friis:
Allows to convert ontologies between different file formats. 

It would be great to add an option for reasoning too.


francescalb (approved):
Works correctly with EMMO branch 1.0.0-beta. I added some documentation. Please review this before merging

jesper-friis (commented):
Looks good. The reason for the failing tests is because libpoppler73 cannot be found in the ubuntu repository.

francescalb:
Test seems to work now in branch tolerant-labels, so will work here too after merging them .

Issue #66, opened 2020-12-08
Unable to use the vertical interoperability demo .py files


Lexachoc:
Hello, 

I'm using Pycharm to run the codes, my OS is win10, and the Python version is 3.8.3.
When running the "define_ontology.py" file, it shows the error message:

> Traceback (most recent call last):
>   File "C:/Users/Choc/PycharmProjects/pythonProject/define_ontology.py", line 196, in <module>
>     class WeldedComponent(emmo.Component):
>   File "C:\Users\Choc\anaconda3\lib\site-packages\emmo\ontology.py", line 110, in __getattr__
>     attr = self.get_by_label(name)
>   File "C:\Users\Choc\anaconda3\lib\site-packages\emmo\ontology.py", line 420, in get_by_label
>     raise NoSuchLabelError('Ontology "%s" has no such label: %s' % (
> emmo.ontology.NoSuchLabelError: Ontology "emmo-inferred" has no such label: Component

I actually have no idea how to deal with that, so I comment out the "class WeldedComponent(emmo.Component):" part first in order to make it runs well. Then, I keep running the "plot_ontology.py" file, but it still shows error message no matter the  "define_ontology.py" file runs well or not:

> Traceback (most recent call last):
>   File "C:/Users/Choc/PycharmProjects/pythonProject/plot_ontology.py", line 17, in <module>
>     graph = onto.get_graph(entities=classes, relations='all',
>   File "C:\Users\Choc\anaconda3\lib\site-packages\emmo\ontology.py", line 672, in get_graph
>     return OntoGraph(self, **kwargs)
>   File "C:\Users\Choc\anaconda3\lib\site-packages\emmo\graph.py", line 227, in __init__
>     self.add_entities(entities=entities, relations=relations,
>   File "C:\Users\Choc\anaconda3\lib\site-packages\emmo\graph.py", line 242, in add_entities
>     self.add_nodes(entities, nodeattrs=nodeattrs, **attrs)
>   File "C:\Users\Choc\anaconda3\lib\site-packages\emmo\graph.py", line 303, in add_nodes
>     self.add_node(name, nodeattrs=nodeattrs, **attrs)
>   File "C:\Users\Choc\anaconda3\lib\site-packages\emmo\graph.py", line 297, in add_node
>     self.dot.node(label, label=label, **kw)
>   File "C:\Users\Choc\anaconda3\lib\site-packages\graphviz\dot.py", line 130, in node
>     name = self._quote(name)
[...]

Issue #11, opened 2019-07-14
Add semantics for dimensionality to the user case ontology in the demo


jesper-friis:
This will be very useful to e.g. semantically describe that the stiffness_tensor is a 6x6 array and not just 36 real numbers. We can assign a name (and possibly a length) to the dimensions directly in the ontology. This would fit perfect with SOFT/dlite. To make advantage of this, the onto2meta.py module should be updated.

PR #67, opened 2020-12-08
Added a quick fix for vertical demo.

jesper-friis:
Reverted to EMMO version  1.0.0-alpha2.

The demo needs a throughout cleanup - this is just a quick fix such that the vertical demo runs...

Relates to issue #11 and #66


jesper-friis:
Hmm, the pytest fails. Would like to fix that before merge.

PR #68, opened 2020-12-10
Use rdflib to load non-supported formats.

jesper-friis:
This should enable emmocheck and other tools to read turtle out of the box


francescalb (approved):
This works nicely for .ttl ontologies downloaded on the computer. Getting the .ttl directly from emmo-repo.github.io does not work. We can make this as a separate issue?

PR #71, opened 2020-12-11
Fix bug introduced in ontoconvert

jesper-friis:
A bad one - without importing Graph ontoconvert always crash


francescalb (approved):
Reinstalled everything on new machine to make sure I didn't use the wrong tool for some reason. 
This now works correctly.  I suggest making a b-release of this version.

PR #73, opened 2020-12-13
Added sync_python_names() and corrected handling of individuals in sync_attributes()

jesper-friis:
These are needed for generation of the EMMO periodic table domain ontology in https://github.com/emmo-repo/EMMO/pull/143


francescalb (approved):
Works well for me

PR #77, opened 2020-12-22
Fixes needed for access ontologies

jesper-friis:


francescalb (approved):
Documentation of .ttl and ontology from access now works

PR #85, opened 2020-12-22
Automatically find roots in ontograph 

francescalb:
Automatically identify roots if not given as input in ontograph.


jesper-friis (approved):
Looks good

PR #94, opened 2020-12-30
Changed the sync_reasoner() method to use FaCT++ as the default reasoner

jesper-friis:
Add FaCT++ to supported reasoners accessible with sync_reasoner().


jesper-friis:
Reading the reasoned ontology from an rdflib graph back into the owlready2 triplestore needs to be implemented...

francescalb (approved):
Reasoning of emmo within EMMO-python works well.

PR #96, opened 2021-01-03
Require reflib 5.0.0 or later for correct turtle serialisation

jesper-friis:
rdflib prior to 5.0.0 serialises real literals syntactically wrong in turtle.
See https://github.com/RDFLib/rdflib/issues/1043.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,6 +4,6 @@ graphviz
 PyYAML
 blessings
 Pygments
-rdflib
+rdflib>=5.0.0
 semver
 pydot


jesper-friis:
rdflib 5.0.0 doesn't serialise correct. You have to use master or wait for the next rdflib release - add a comment about this.

Add a section "Known issues" in README file where we describe this

francescalb:
No reason to do this as it does not solve the issue. The problem is serializing numbers going from owl-> ttl. This is currently not a part of our workflow. The issue is fixed in the next release of rdflib (6.x.y). Just add a comment to readme about this so it is not forgotten.

jesper-friis:
Agree, we close the issue

PR #102, opened 2021-01-04
Allow to load turtle ontologies without catalog file.

jesper-friis:
Before convert_imported() was always looking for a catalog file and would
fail if it couldn't be found.  Solved by added `url_from_catalog` option
to convert_imported().


francescalb (approved):
Tested

PR #106, opened 2021-01-04
Updated pandoc req in Dockerfile

francescalb:
Updated the Dockerfile with correct pandoc requirements, and installs EMMO-python with correct python packages automatically from requirements. 

I have not updated the development version. Do we still need it?


Dockerfile
```diff
@@ -1,13 +1,12 @@
 from continuumio/miniconda3
-
-RUN apt update && apt install -y texlive-latex-extra pandoc gcc gcj-jre
-RUN conda install -c conda-forge graphviz
-RUN pip install Cython pydot ase Owlready2
+RUN apt update && apt install -y texlive-latex-extra texlive-xetex graphviz
```
jesper-friis, line 2: I get the following error

    E: Unable to locate package gcj-jre

It may be due to the package being temporarily unavailable in the apt repo

francescalb: That is very strange since we are no longer asking for that package, at least not explicitly. It works for me,  can you try again?

jesper-friis: My fault, I was in the wrong branch. It builds fine. 

My only question is what we are going to use the docker image for? I can think about two use cases:
1) to provide the right version of pandoc, graphviz and other non-Python dependencies
2) to make it simple and fast to use the tools from EMMO-python in docker-based automated testing

Either use case would require some additional work

jesper-friis (approved):

Dockerfile
francescalb: 1. With this Dockerfile I am able to get the right versions of graphviz and genereate the pdf documentation that I am not otherwise able to generate on the Centos7

PR #107, opened 2021-01-05
Updated the Ontology.sync_attributes() method.

jesper-friis:
Added `class_docstring` argument to sync_attributes() and made the uuid be uniquely determined by the class name.

The latter ensures that the same source file always produces the same result.


francescalb (approved):
Looks good

PR #110, opened 2021-01-09
Break recursion error in get_by_label()

jesper-friis:


francescalb (approved):
Nicely spotted! works well

PR #113, opened 2021-01-17
Update emmocheck exceptions

jesper-friis:
Also allowed to specify ontologies to skip in emmocheck configuration file.


francescalb (approved):
Very good, I updated the version so that a new pip-release can be made ready for BIGMAP

PR #114, opened 2021-01-17
WIP: Improve load ontology

jesper-friis:
work in progress - should be better tested


jesper-friis:
Already merged into master

PR #116, opened 2021-01-18
Update publish workflow + add dependabot

### LINKED ISSUE #115: Update publish workflow
Utilize the [dedicated PyPI publish action](https://github.com/pypa/gh-action-pypi-publish) and run only on published GitHub releases.

CasperWA:
This PR closes #115 by updating the publish to PyPI workflow. Note, the workflow will only run if a GitHub release is _published_. I.e., one can create draft releases and hold off publishing it until it's relevant.

It's worth noting that the built distributions (bdist) rely on the OS, here being Ubuntu 20.04. If a specific OS is desired, the `runs-on` value should be changed accordingly.

This PR also adds dependabot, which will run for two different cases:
- **pip**
  Monday mornings, it will run to check if there are newer versions released for the packages specified in `requirements.txt`, comparing to the versions specified in said file. Since only a single package has a version specified (Owlready2), it will be the only one targeted at this point in time.
- **actions**
  Every weekday morning, it will check whether there are newer versions for the actions used in the GitHub Actions workflows.

Finally, I've updated the used actions in both workflows to the latest.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+version: 2
+updates:
+- package-ecosystem: pip
+  directory: "/"
+  schedule:
+    interval: weekly
+  # Should be bigger than or equal to the total number of dependencies (currently 9)
+  open-pull-requests-limit: 15
+  target-branch: master
+- package-ecosystem: github-actions
+  directory: "/"
+  schedule:
+    interval: daily
+  target-branch: master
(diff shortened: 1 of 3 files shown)


CasperWA:
Ah. You might also want to consider using an API token for PyPI instead of storing your username and password in the project secrets. Then the `username` and `password` should be changed to:
```yml
        user: __token__
        password: ${{ secrets.PYPI_TOKEN }}
```

jesper-friis (approved):
I added the PYPI_TOKEN as suggested.

jesper-friis:
Looks fine.

PR #126, opened 2021-02-27
Use catalog by default

jesper-friis:


emmo/utils.py
```diff
@@ -6,10 +6,20 @@
...
 
 import owlready2
 
 
+# Format mappings: file extension -> rdflib format name
+FMAP = {
+    'n3': 'ntriples',
+    'nt': 'ntriples',
+    'ttl': 'turtle',
+    'rdfxml': 'xml',
```
francescalb, line 19: Should we not also add .owl as possible file extension?

jesper-friis: rdflib already recognises the .owl extension as the xml format, so it is not needed

francescalb (approved):
Approved. I have changed default in get_ancestors to 'all' and updated the documentation

PR #139, opened 2021-03-15
Fixed reading xml as 'rdfxml'

francescalb:
Note that owlxml is not supported now...

### DIFF
--- a/emmo/ontology.py
+++ b/emmo/ontology.py
@@ -21,7 +21,6 @@ from rdflib.util import guess_format
 
 import owlready2
 from owlready2 import locstr
-from owlready2.driver import _guess_format
 
 from .utils import asstring, read_catalog, infer_version, convert_imported
 from .utils import FMAP, isinteractive, ReadCatalogError
--- a/emmo/ontology.py
+++ b/emmo/ontology.py
@@ -392,7 +392,8 @@ class Ontology(owlready2.Ontology, OntoGraph):
                 return super().load(only_local=only_local,
                                     reload=reload,
                                     reload_if_newer=reload_if_newer,
-                                    format=fmt,
+                                    format='rdfxml' if fmt == 'xml' 
+                                                    else fmt,
                                     **kwargs)
 
             else:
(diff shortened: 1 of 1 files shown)


francescalb:
> Note that owlxml is not supported now...

I think it all works now, so please review

emmo/ontology.py
```diff
@@ -69,7 +69,7 @@ def get_ontology(self, base_iri='emmo-inferred'):
             of EMMO
         """
         if base_iri == 'emmo':
-            base_iri = 'http://emmo.info/emmo'
+            base_iri = 'https://emmo-repo.github.io/latest-stable/emmo.owl'
```
jesper-friis, line 72: By loading EMMO as a single file we will loose the possibility to easily list the individual ontologies separately as well as the documentation of the individual ontologies. On the positive side, it will load much faster. It is up to us to decide. Probably this is a good choice.

emmo/ontology.py
```diff
@@ -69,7 +69,7 @@ def get_ontology(self, base_iri='emmo-inferred'):
...
         elif base_iri == 'emmo-inferred':
             base_iri = (
                 'https://emmo-repo.github.io/latest-stable/emmo-inferred.ttl')
```
jesper-friis, line 75: Probably a good idea to load the owl file instead of ttl to save temporary conversion and storage of the ontology.

jesper-friis (commented):
Suggest that "emmo-inferred" should load the owl file instead of ttl

emmo/ontology.py
francescalb: OK, done. Have to change this if emmo eventually stops supporting owl

emmo/ontology.py
francescalb: I think importing separate files is the most correct for this. Otherwise the inferred version can be uploaded. I changed the uri to the raw as http://emmo.info/emmo did not work

jesper-friis (approved):

PR #142, opened 2021-03-18
Corrected configuration of exceptions for test_class_label test.

jesper-friis:
This should fix the failing tests in domain-atomistic


francescalb (approved):
This is fine

PR #144, opened 2021-03-21
Save to turtle and ontology annotations (via the metadata attribute)

jesper-friis:
Added some new features (used for generating the cif ontology):
  - save to turtle and other formats supported by rdflib
  - added more methods to ontology metadata: keys(), items(), ++
  - implemented annotate_with_ontology() utility function

### DIFF
--- a/emmo/patch.py
+++ b/emmo/patch.py
@@ -1,8 +1,11 @@
 # -*- coding: utf-8 -*-
 """This module injects some additional methods into owlready2 classes."""
+import types
 
 import owlready2
 from owlready2 import ThingClass, PropertyClass, Thing, Restriction, Namespace
+from owlready2 import Metadata
+from owlready2 import rdf_type, owl_annotation_property
 
 
 # Improve default rendering of entities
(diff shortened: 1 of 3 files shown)


emmo/patch.py
```diff
@@ -176,23 +201,79 @@ def namespace_init(self, world_or_ontology, base_iri, name=None):
...
 
-# Method names for individuals must be different from method names for classes
-type.__setattr__(Thing, 'get_preflabel', get_preferred_label)
-type.__setattr__(Thing, 'get_individual_annotations',
-                 get_individual_annotations)
+
+#
+# Extending Metadata
+# ==================
+def keys(self):
```
francescalb, line 210: Does this take the ontology as argument? emmo.ontology.Ontology does not have .namespace.

emmo/utils.py
```diff
@@ -420,7 +424,7 @@ def infer_version(iri, version_iri):
     return version
 
 
-def ontology_annotation(onto, imported=True):
+def annotate_with_ontology(onto, imported=True):
```
francescalb, line 427: This functions seems to work well, should it perhaps be added as option to Ontology.save?

jesper-friis: That could also be a possibility. I original intention was to add it as an option to the ontoconvert tool.

francescalb (approved):

PR #145, opened 2021-04-01
WIP: Issue128 return ontology.namespace.entity

francescalb:


emmo/ontology.py
```diff
@@ -185,6 +189,19 @@ def get_by_label(self, value, label_annotations=None):
...
+            try:
+                return self.get_by_label_all(
+                            value,
+                            label_annotations=label_annotations,
+                            namespace=namespace)[0]
+            except IndexError:  # Should ad secial exception here
+                raise NoSuchLabelError('No label annotations matches %s'
+                                       'in namespace %s' % (value, namespace))
+
+        if value in self.namespaces:
```
jesper-friis, line 202: This may result in infinite recursion, in the case we are creating a new ontology (and hence not calling `load()`), since accessing self.namespaces will then call `__getattr__()` which in turn will call `get_by_label()`. One way to avoid this is to replace line 202 with:

    if 'namespaces' in self.__dict__ and value in self.namespaces:

Hmm, there is another issue. New namespaces added after we have called `load()` will not be in `self.namespaces`. I guess that we want to cache the namespaces in a dict for performance reasons. A possible solution could be to update `self.namespaces` in the `sync_attributes()` method. What do you think?

francescalb:
Most of this has been added a little bit differently, with prefix
so that all ontologies have a prefix and get_by_label accepts "prefix:label" as argument.

ontology.prefix.label is not supported and we have decided to not do that for now.

PR #146, opened 2021-04-07
Improved issue reporting in emmocheck

jesper-friis:


francescalb (approved):
Looks good. The 'is' test is not being used in EMMO-1.0.0beta, is there an ontology where you have tested this?

PR #151, opened 2021-04-26
Issue150 ontograph cannotload emmo inferred directly

francescalb:
ontograph [options] emmo-inferred did not work. Added --no-catalog-option as in emmocheck. Please check.

### DIFF
--- a/tools/ontograph
+++ b/tools/ontograph
@@ -45,6 +45,10 @@ def main():
         'ontology.  This option is used together with --local and '
         'defaults to "catalog-v001.xml".')
     parser.add_argument(
+        '--no-catalog', action='store_false', dest='url_from_catalog',
+        default=None,
+        help='Whether to not read catalog file even if it exists.')
+    parser.add_argument(
         '--path', action='append', default=[],
         help='Paths where imported ontologies can be found.  May be provided '
         'as a comma-separated string and/or with multiple --path options.')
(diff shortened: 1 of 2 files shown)


tools/ontograph
```diff
@@ -44,15 +44,20 @@ def main():
...
+        '--no-catalog', action='store_false', dest='url_from_catalog',
+        default=None,
+        help='Whether to not read catalog file even if it exists.')
     parser.add_argument(
         '--path', action='append', default=[],
         help='Paths where imported ontologies can be found.  May be provided '
         'as a comma-separated string and/or with multiple --path options.')
     parser.add_argument(
-        '--reasoner', nargs='?', const='HermiT', choices=['HermiT', 'Pellet'],
+        '--reasoner', nargs='?', const='HermiT',
```
jesper-friis, line 56: We should make FaCT++ the default reasoner: const='FaCT++'

jesper-friis (commented):
No problem to load inferred EMMO. We should just set FaCT++ to the default reasoner if --reasoner is provided without argument: Replace "const='HermiT'" with "const='FaCT++'".

francescalb:
> No problem to load inferred EMMO. We should just set FaCT++ to the default reasoner if --reasoner is provided without argument: Replace "const='HermiT'" with "const='FaCT++'".

Fixed

jesper-friis (approved):

PR #156, opened 2021-04-27
Periodic table example

francescalb:


examples/periodic-table/periodic-table.py
```diff
@@ -0,0 +1,133 @@
...
+# define function that returns a string in English in owlready2
+def en(s):
+    """Returns `s` as an English location string."""
+    return owlready2.locstr(s, lang='en')
+
+
+# Load emmo
+world = World()  # to make it extensible for when an ontology gets very large
+emmo = world.get_ontology(
+        'https://raw.githubusercontent.com/emmo-repo/EMMO/periodic-table/'
```
jesper-friis, line 17: This url is not resolved when opening the generated ontology in Protege. We should probably use 
https://emmo-repo.github.io/versions/1.0.0-beta/emmo.ttl instead.

examples/periodic-table/periodic-table.py
```diff
@@ -0,0 +1,133 @@
...
+    'Periodic table is released under the Creative Commons Attribution 4.0 '
+    'International license (CC BY 4.0).'))
+
+onto.metadata.title.append(en('Periodic table'))
+onto.metadata.creator.append(en('Jesper Friis'))
+onto.metadata.contributor.append(en('SINTEF'))
+onto.metadata.creator.append(en('Emanuele Ghedini'))
+onto.metadata.contributor.append(en('University of Bologna'))
+onto.metadata.publisher.append(en('EMMC ASBL'))
+onto.metadata.license.append(en(
```
jesper-friis, line 114: onto.metadata.license is not resolved when opening in Protege

jesper-friis (commented):
I am little unsure about replacing getlabel() with asstring() even thought it should work with the current implementation. They kind of have slightly different purpose. The first returns the label of an entity, the latter a string representation.

francescalb:
> I am little unsure about replacing getlabel() with asstring() even thought it should work with the current implementation. They kind of have slightly different purpose. The first returns the label of an entity, the latter a string representation.

I agree, it has been changed to get_label and fixed in another PR.

jesper-friis (approved):
Looks very good!

PR #159, opened 2021-04-30
Added write_catalog()

jesper-friis:


francescalb (approved):
Very good, we might want to add a write_catalog-option to ontology.save?

PR #162, opened 2021-05-07
Made it possible to load other ontologies like foaf

jesper-friis:
Common ontologies where we refer to entities by name rather than skos:prefLabel or rdfs:label are badly supported by EMMO-python. This PR make it possible to load foaf:

```python
from emmo import get_ontology
skos = get_ontology('http://www.w3.org/2004/02/skos/core#').load()
foaf = get_ontology("http://xmlns.com/foaf/0.1/")
foaf.imported_ontologies.append(skos)
foaf._special_labels = ()  # <-- needed to not get confused by labels containing spaces, not convenient!
foaf.load()

emmo = get_ontology().load()

with emmo:
    class Person(emmo.Interpreter):
        equivalent_to = [foaf.Person]
```

We should think about a more convenient solution

### DIFF
--- a/emmo/ontology.py
+++ b/emmo/ontology.py
@@ -299,7 +299,7 @@ class Ontology(owlready2.Ontology, OntoGraph):
                    tmpdir=tmpdir, **kwargs)
 
         # Enable optimised search by get_by_label()
-        if not self._special_labels:
+        if self._special_labels is None:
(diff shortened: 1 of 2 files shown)


francescalb (changes requested):
Can you add a test importing foaf   (or something else) into emmo, and checking that getting entities from there works?

francescalb (approved):
Good, thank you

PR #163, opened 2021-05-07
WIP: Ipycytoscape

francescalb:
Adding ipycytoscape function to graph.py.


francescalb:
the ipycytoscapegraph is now working with arrow styles similar to those from OntoGraph. Not that inverse arrows are 'diamonds' instead of inverse colours. Correct colours for the classes according to OntoGraph has not been implemented (yet). 

The styles are now defined in a quite cumbersome way, so help to improve this is greatly appreciated. The style information is all in the OntoGraph.dot.source , but this info is somehow lost in the conversion to networkx which goes throgh pydotplus.

jesper-friis (commented):
Very nice. I guess we can remove the jpg images of Bjørn Tore and Sylvain.

jesper-friis (approved):

PR #167, opened 2021-05-14
Owlready requirement >0.28

francescalb:
Please check all our packages and see if they are within the required range


requirements.txt
```diff
@@ -1,9 +1,9 @@
...
-pydot
+Cython~=0.29.21
+Owlready2~=0.28
+graphviz~=0.16
+PyYAML~=5.4.0
+blessings~=1.7
+Pygments~=2.8.0
+rdflib~=5.0.0
+semver~=2.13.0
+pydot~=1.4.2
```
jesper-friis: Should be removed together with the deprecated emmo/ontograph.py
We should make a new issue for that.

jesper-friis (approved):

PR #169, opened 2021-05-21
WIP: Added namespaces dict and namespace support in get_by_label

francescalb:
ontology.namespaces is now included as a dict with short name as key and iri as value.
ontology.get_by_label(label, namespace='namespace_shortname') is now supported.
same for ontology.get_by_label_all


emmo/ontology.py
```diff
@@ -150,6 +151,18 @@ def __dir__(self):
         if self._dir_name:
             s.update(e.name for e in lst if hasattr(e, 'name'))
 
+        # updateing with namespace keys
+        # Attention: What to do if a prefLabel coincides with a namespace key?
+        if 'namespaces' not in self.__dict_:
```
jesper-friis, line 156: Nice. We repeat this several places. Probably a good idea to factor it out into an own method.

jesper-friis: An alternative would be to make 'namespaces' a property. Then 'namespaces' would always be there and behave as an attribute, but will under the hood be created the first time it is accessed.

francescalb: > An alternative would be to make 'namespaces' a property. Then 'namespaces' would always be there and behave as an attribute, but will under the hood be created the first time it is accessed.

NOt sure I follow.
like @property and def namespace(

francescalb:
Most of this has been added a little bit differently, with prefix
so that all ontologies have a prefix and get_by_label accepts "prefix:label" as argument.

PR #177, opened 2021-06-11
Fixed import of defaultstyle in ontograph-tool

francescalb:
Default style in in graph.py has been moved out of OntoGraph previously, so as to serve both OntoGraph and cytoscapegraph.

Here import of _default_style in the ontograph tool is fixed


jesper-friis (approved):
Enkel og grei

PR #204, opened 2021-08-22
ttl standard for emmo

francescalb:
Made loading of ttl version standard for emmo, emmo-inferrred and development version.
Corrected test_catalog (change search for ttl and not owl in catalog) and test_load accordingly (added owl version for testing of owl)

### DIFF
--- a/emmo/tests/test_load.py
+++ b/emmo/tests/test_load.py
@@ -9,7 +9,7 @@ from emmo import get_ontology  # noqa: E402, F401
 
 
 # Check that the defaults works
-emmo = get_ontology('emmo').load()  # owl format
+emmo = get_ontology('emmo').load()  # ttl format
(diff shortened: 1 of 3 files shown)


emmo/tests/test_load.py
```diff
@@ -18,6 +18,11 @@
 emmo = get_ontology('emmo-development').load()  # ttl format
 assert emmo.Atom.prefLabel.first() == 'Atom'
 
+emmo = get_ontology('https://emmo-repo.github.io/latest-stable/'
+                    'emmo-inferred.owl').load()  # owl format
```
jesper-friis, line 22: Should we also switch to ttl here? Both owl and ttl is available.

jesper-friis (approved):
Looks very good

PR #207, opened 2021-09-03
Added function new_entitiy to ontology

francescalb:


emmo/ontology.py
```diff
@@ -1014,3 +1015,12 @@ def get_wu_palmer_measure(self, cls1, cls2):
         n1 = self.number_of_generations(cls1, cca)
         n2 = self.number_of_generations(cls2, cca)
         return 2 * ccadepth / (n1 + n2 + 2 * ccadepth)
+
+    def new_entity(self, name, parent):
```
jesper-friis, line 1019: Now I am picky. PEP8 suggests that docstrings uses 3 double-quotes and that there are no newline after the initial quotes. See https://www.python.org/dev/peps/pep-0008/#documentation-strings

Example:

    """Return a foobang
    
    Optional plotz says to frobnicate the bizbaz first.
    """

jesper-friis (approved):
Just a style comment, otherwise it looks very good.

PR #211, opened 2021-09-15
Avoid using Owlready2 v0.34

### LINKED ISSUE #210: Circular reference from Owlready2
A `sqlite3.OperationalError: circular reference: prelim1_objs` is thrown when running `DocPP.process()`.
This error is thrown from Owlready2 and doesn't seem to be directly related to this package, but never the less affects it severely.

An example of the error message can be seen in [this CI run](https://github.com/emmo-repo/EMMO-python/pull/208/checks?check_run_id=3599727802#step:6:28)

CasperWA:
Fixes #210

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 Cython>=0.29.21
-Owlready2>=0.28,!=0.32
+Owlready2>=0.28,!=0.32,!=0.34
 graphviz>=0.16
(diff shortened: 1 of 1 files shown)


francescalb (approved):
Good, if this problem persists in the next version of Owlready2, we should check it out more closely

PR #214, opened 2021-09-15
Remove sub-fact++ modules importability

### LINKED ISSUE #213: Remove importability of sub-`factpluspluswrapper` folders
Sub-`factpluspluswrapper` folders do not need to be importable in Python, since these are relevant only for Java.

CasperWA:
I.e., remove them as Python importable modules.
Ensure all files are still included in the package build.

Closes #213.


francescalb (changes requested):
Good, but I suggest adding a test that uses the reasoner. in test_basic there is a commented out emmo.sync_reasoner(), commented out before we got Fact++ I think. Can we uncomment that one or will the tests be too time-consuming then?

CasperWA:
This has been superseded by #217.

PR #216, opened 2021-09-21
Update requirements

### LINKED ISSUE #206: Require rdflib>5.0.0?

CasperWA:
~This does not close the #206 issue anymore. See the [comment below](https://github.com/emmo-repo/EMMO-python/pull/212#issuecomment-921229187).~
Closes #206.
This now closes #206, because it has been decided to keep support for `rdflib<6.0.0` in order to support system-wide installation and use of EMMOntoPy. With the inclusion of the warnings here, we also let users know that if they want to properly output turtle files (for certain edge-cases) they should ensure they can install the latest `rdflib` version.

Set version lofts for Python dependencies.

Note, this is a remake of #212, but using a branch from this repository instead of a fork.


CasperWA:
@francescalb This has been remade from my forked branch, so it should be easier to test locally now without having to add my fork as a remote :)

francescalb (changes requested):
Should be included in ontology.save as well. Now saving a new ontology in .ttl format with rdflib 4.2.2 does not return an error.

francescalb:
I am a little unsure if it is not best to give a Warning rather than an error. It depends on the scientific notation is commonly used or not.

CasperWA:
> I am a little unsure if it is not best to give a Warning rather than an error. It depends on the scientific notation is commonly used or not.

Fine by me.
Especially after looking at the comment in the README. It indeed seems to be for specific cases.

francescalb:
The whole warning when saving is not very pretty, but maybe it should be like that?
>>>from ontopy import get_ontology
>>>o = get_ontology('emmo').load()
>>> o.save('ontology.ttl')
/full/path/to/EMMO-python/ontopy/ontology.py:506: IncompatibleVersion: To correctly convert to Turtle format, rdflib must be version 6.0.0 or greater, however, the detected rdflib version used by your Python interpreter is '4.2.2'. For more information see the 'Known issues' section of the README.
  warnings.warn(
>>>

CasperWA:
> The whole warning when saving is not very pretty, but maybe it should be like that?

This is the standard Python warning message formatting. The formatting can be changed, however, this should be subject of a separate PR I think, otherwise this particular PR will be bogged down with warning formatting discussions and not with a discussion regarding the requirements, as it should.

I suggest you create an issue from your comment and maybe outline how you would like it to be formatted. Or just at least create the issue, then we can discuss it later :)

francescalb (approved):
Now works very nicely.

CasperWA:
As a note, I've made the warning "pretty" for the CLI tool :)

francescalb:
> As a note, I've made the warning "pretty" for the CLI tool :)

I noticed :)

PR #218, opened 2021-09-21
Unskip test as #210 has been resolved

CasperWA:


francescalb (approved):
Good, if it comes up again with newer versions of Owlready2 we need to consider it further.

Issue #199, opened 2021-08-18
Documentation for writing documentation


CasperWA:
Write a wiki page or a an extra page for developers and contributors on style guide and style rules we follow to write documentation.

PR #222, opened 2021-09-21
Setup new documentation framework

### LINKED ISSUE #193: Clean up demo, examples and docs
Move `demo` and `examples` under `docs` and clean up the structure surrounding it.

### LINKED ISSUE #195: Setup Materials for MkDocs framework
- Requirements
- Standard/initial ToC (table of contents)
- GH Actions autobuild & release
- GH Actions CI build checks

### LINKED ISSUE #197: Consider using `mike` for versioned documentation
[`mike`](https://github.com/jimporter/mike).

CasperWA:
Closes #193.
Closes #195.

This PR sets up a MkDocs framework.

All markdown files have been updated to follow a better `git diff` strategy with "one-sentence-per-one-line". While it's not a perfect strategy, with obvious shortcomings for "thin" terminals, it still succeeds in bringing much better `git diff`s than otherwise, where reviewing may become so difficult mistakes are made due to obscurity.

A `docs` extra is added, meaning one can now install all the Python dependencies needed to build the documentation as well as run a simple server locally by running `pip install -e .[docs]` (in the future this will be `pip install EMMOntoPy[docs]`.

[Invoke](http://www.pyinvoke.org/) tasks have been created and added to a root Python file `tasks.py` to auto-generate the Python API reference documentation, as well as update the root `index.md` file in the documentation from the root `README.md` - changing relative links to match, etc.
These tasks can be used for continuous deployment (CD) of the documentation.

Another task has been added to auto-update the version, where needed, in the code.
This means one can use this in CD of the package when releasing a new version.

I've also added the logo from the Settings page as the "favicon" and general logo of the documentation.

As a note, in order to close #193, we need to decide what should happen with the root folders `demo` and `examples`.
According to the referenced issue, it is suggest to move them under `docs`. In the current PR I have made symbolic links to them for now, but this might as well be settled in this PR together with the setup of the documentation framework.

---

To test this out, you can checkout this branch ([`cwa/close-195-mkdocs-framework`](https://github.com/emmo-repo/EMMO-python/tree/cwa/close-195-mkdocs-framework)) and run:

```console
$ pip install -U pip setuptools wheel
Obtaining ...
Successfully installed ...
$ pip install -U -e .[docs]
Obtaining ...
Successfully installed ...
$ mkdocs serve
INFO     - Building documentation ...
...
INFO     - [...] Serving on http://127.0.0.1:8000/EMMO-python/
```

Then go to the address it prints to your terminal and poke around.

---

WIth the latest inclusions this now also closes #197.
For more information see [this comment](https://github.com/emmo-repo/EMMO-python/pull/222#issuecomment-924999036) below.


CasperWA:
An example of using versioned documentation can be seen by doing the following:

```console
$ cd /path/to/EMMO-python
$ git fetch origin
$ git checkout test-mike
$ pip install -e .[docs]
$ mike serve -b test-mike
Starting server at http://localhost:8000/
Press Ctrl+C to quit.
```

Now go to [http://localhost:8000](http://localhost:8000) and see how the versioned documentation will look.
Essentially it builds the documentation for each specified version and puts it in a folder (see the file structure of the [test-mike branch](https://github.com/emmo-repo/EMMO-python/tree/test-mike)) and then one can switch between these versions through the dropdown at the top.
I've created some test version here, it's all the same documentation, but the system doesn't know that :) It defaults to the "latest" version, which I've set as an alias for 1.0.3.

If this is desirable, we can implement this type of documentation for the GitHub pages. However, I am unsure whether we shouldn't actually use `ontodoc` to create the whole documentation instead, or? It seems weird to not use the tools of the package - although, I guess that tool is for creating documentation for ontologies and not Python packages 😉 
In any case, it might be good to have the output of using `ontodoc` as an example to be able to show in the documentation. I'll think about how this can be done.

francescalb:
> An example of using versioned documentation can be seen by doing the following:
> 
> ```
> $ cd /path/to/EMMO-python
> $ git fetch origin
> $ git checkout cwa/close-195-mkdocs-framework
> $ pip install -e .[docs]
> $ mike serve -b test-mike
> Starting server at http://localhost:8000/
> Press Ctrl+C to quit.
> ```
> 
> Now go to http://localhost:8000 and see how the versioned documentation will look.
> Essentially it builds the documentation for each specified version and puts it in a folder (see the file structure of the [test-mike branch](https://github.com/emmo-repo/EMMO-python/tree/test-mike)) and then one can switch between these versions through the dropdown at the top.
> I've created some test version here, it's all the same documentation, but the system doesn't know that :) It defaults to the "latest" version, which I've set as an alias for 1.0.3.
> 
> If this is desirable, we can implement this type of documentation for the GitHub pages. However, I am unsure whether we shouldn't actually use `ontodoc` to create the whole documentation instead, or? It seems weird to not use the tools of the package - although, I guess that tool is for creating documentation for ontologies and not Python packages 😉
> In any case, it might be good to have the output of using `ontodoc` as an example to be able to show in the documentation. I'll think about how this can be done.

I installed mike with `pip install mike`, but still get: mike serve -b test-mike
Config value: 'theme'. Error: Unrecognised theme name: 'material'. The available installed themes are: readthedocs, mkdocs
Config value: 'markdown_extensions'. Error: Failed loading extension "pymdownx.highlight".
Config value: 'plugins'. Error: The "mkdocstrings" plugin is not installed
error: Aborted with 3 Configuration Errors!

Can you include documentation on what is needed to make it locally?

CasperWA:
> I installed mike with `pip install mike`, but still get: mike serve -b test-mike
> Config value: 'theme'. Error: Unrecognised theme name: 'material'. The available installed themes are: readthedocs, mkdocs
> Config value: 'markdown_extensions'. Error: Failed loading extension "pymdownx.highlight".
> Config value: 'plugins'. Error: The "mkdocstrings" plugin is not installed
> error: Aborted with 3 Configuration Errors!
> 
> Can you include documentation on what is needed to make it locally?

Yes, you need to follow my instructions and install the package with the `docs` extra :)
Either that or you can do it more "manually" by installing the requirements for the documentation directly from the `requirements_docs.txt` file, i.e.: `pip install -r requirements_docs.txt`.

francescalb:
That is what i did: pip install -e .[docs]

(as per your instructions) and the error message reported is what I got.

Manual setup works, but the instructions as written do not.

CasperWA:
> That is what i did: pip install -e .[docs]
>
> (as per your instructions) and the error message reported is what I got.
>
> Manual setup works, but the instructions as written do not.

Fair enough 😅 Sorry for the apparent badgering then. I'll check out why it doesn't work, it should 😕 oh well 😄

francescalb:
I think this looks very very nice! As for the docs and demo, I am now convinced that we should only have those in the documentation :) We should definitely have documentation on ontodoc, whic is supposed to be for documenting ontologies and not python packages as you already wrote. Is there a way to make binders with this setup for e.g. a jupyter notebook example?

CasperWA:
> I think this looks very very nice!

Yay :)
It's using the Materials theme for MkDocs (see [here](https://squidfunk.github.io/mkdocs-material/)).
Which I think makes it look awesome :)
But good that this work was not in vain.

> As for the docs and demo, I am now convinced that we should only have those in the documentation :)

Okay - I'll try to make them a bit nicer then.
I'm still unsure what the different parts of these examples and demos are - like, is the example for `emmodoc` actually used for the EMMO documentation or is it just an early copy of what is in the EMMO documentation?

> We should definitely have documentation on ontodoc, whic is supposed to be for documenting ontologies and not python packages as you already wrote. Is there a way to make binders with this setup for e.g. a jupyter notebook example?

Hmm. I'll see what I can do. Mybinder is quite slow to start up if it hasn't been run in a while, and therefore cached, but it's a good thing to have, at least.

francescalb:
to clarify, I think we are more or less at the point that we can make the first release and then make other additions as improvements

CasperWA:
This is ready for review. It has been greenlit by @francescalb (most likely following up with an "Approved" review), but it would be good to get @jesper-friis's input before merging this in.

Note, on merge this will NOT start publishing new documentation. The existing documentation (see https://emmo-repo.github.io/EMMO-python) will continue as is, one needs to (for now) manually publish this documentation if one wants to. Automating and documenting this is partly part of issue #199. However, I will create a dedicated issue for this.

CasperWA:
Argh, it seems the `docs` and `examples` folders were included in the package build. While I've now updated all references to the folders, I don't see why these need to be included in the build? I'd argue for removing them and keeping them only for use with the (new) documentation and for people who clone the repository from GitHub.

CasperWA:
> Argh, it seems the `docs` and `examples` folders were included in the package build. While I've now updated all references to the folders, I don't see why these need to be included in the build? I'd argue for removing them and keeping them only for use with the (new) documentation and for people who clone the repository from GitHub.

Concerning this, maybe the last two commits here should be reverted and a dedicated PR could be made, instead of trying to both change the package build as well as setup a documentation framework here?

francescalb:
How about adding the demo and examples here now separately (copy instead of soft link), so that they are in the documentation from the start, and fix the package(eg remove them) in a separate pr? Maybe this is what you suggested?

CasperWA:
> How about adding the demo and examples here now separately (copy instead of soft link), so that they are in the documentation from the start, and fix the package(eg remove them) in a separate pr? Maybe this is what you suggested?

Yeah, this is what I suggested. However, I would keep the soft links for now, as I see no reason to pollute the history with even more MB :)

francescalb:
> > How about adding the demo and examples here now separately (copy instead of soft link), so that they are in the documentation from the start, and fix the package(eg remove them) in a separate pr? Maybe this is what you suggested?
> 
> Yeah, this is what I suggested. However, I would keep the soft links for now, as I see no reason to pollute the history with even more MB :)

OK :)

README.md
```diff
@@ -1,158 +1,132 @@
-EMMO - Python API for the European Materials & Modelling Ontology
-=================================================================
+# EMMO - Python API for the European Materials & Modelling Ontology
```
jesper-friis, line 1: # EMMOntoPy - Python library for working with ontologies
EMMOntoPy is a Python library for woking with ontologies in general and the Elemental Multiperspective Material Ontology in particular.

README.md
```diff
@@ -1,158 +1,132 @@
...
+
+<!-- markdownlint-disable MD033 -->
 
 ![CI tests](https://github.com/emmo-repo/EMMO-python/workflows/CI%20Tests/badge.svg)
 [![PyPI version](https://badge.fury.io/py/EMMO.svg)](https://badge.fury.io/py/EMMO)
 
+This package is based on [Owlready2] and provides an intuitive representation of [EMMO] in Python.
+It is available on [GitHub][EMMOntoPy] and on [PyPI][PyPI:EMMO] under the open source [BSD 3-Clause license](LICENSE.txt).
+
+The European Materials & Modelling Ontology (EMMO) is an ongoing effort to create an ontology that takes into account fundamental concepts of physics, chemistry and materials science and is designed to pave the road for semantic interoperability.
```
jesper-friis, line 11: Elemental Multiperspective Material Ontology

README.md
jesper-friis: I think that the old header style with underline is more readable as text

docs/index.md
```diff
@@ -0,0 +1,165 @@
+# EMMO - Python API for the European Materials & Modelling Ontology
```
jesper-friis, line 1: This looks like a copy of README.md. If possible to refer to the README.md file or copy it for the documentation to have a single source of information.

jesper-friis (approved):
Looks great. Had a few comments

README.md
CasperWA: But that will mix styles, as the old style doesn't support headers at level 3 and on. So for header levels 1 and 2 it would be the underline styling and for sub-headers below that it would be the hash-styling. I find the mixing quite confusing.

If documents just have two header levels it's fine for me, but that then means that we would have mixed styles throughout our documentation.

README.md
CasperWA: Yes, this should be changed everywhere, here I have not changed content, only very minor in specific cases. But I can do a commit that goes through the entire documentation and updates this, but it can be done in a dedicated PR as well.

docs/index.md
CasperWA: This is copied/generated from the `invoke` task `create-docs-index`, changing links to the License page and relative links from the new location. See https://github.com/emmo-repo/EMMO-python/pull/222/files#diff-2e746c9eacad6e2c2fdefdc6a665b5ce4607ea384ebf97a729044b79acc778d6R163

README.md
CasperWA: But in the end, if you feel strongly to keep the original styling in the markdown I guess this is fine - we should just do this all over then. That's the only thing I feel strongly about: Keeping the same style (whether it's mixed or not) all over the documentation.

CasperWA: I've created an issue from this to continue the discussion: #231

README.md
CasperWA: I've created an issue for this to track the update of this: #230

PR #226, opened 2021-09-22
Added function to load the emmo (the ontology) directly

### LINKED ISSUE #209: Make function that automatically loads emmo
Should be
from emmo import emmo
emmo = emmo() 
and emmo() returns get_ontology('emmo-inferred').load()

Might be placed in emmopy when we do the refactoring of the package.

francescalb:
Closes #209.

We could call it get_emmo instead.

### DIFF
--- a/emmopy/emmopy.py
+++ b/emmopy/emmopy.py
@@ -1,0 +1,19 @@
+from ontopy import get_ontology
+from typing import Optional
+
+
+def emmo(inferred: Optional[bool] = True):
+    """Returns the current version of emmo.
+
+    Args:
+        inferred: Whether to import the inferred version of emmo or not.
+        Default is True.
+
+    Returns:
+        the loaded emmo ontology.
+    """
+    if inferred == True:
+        name = 'emmo-inferred'
+    else:
+        name = 'emmo'
+    return get_ontology(name).load()
--- a/emmopy/emmopy.py
+++ b/emmopy/emmopy.py
@@ -15,6 +15,5 @@ def get_emmo(inferred: Optional[bool] = True) -> "Ontology":
     Returns:
         The loaded emmo ontology.
     """
-    inferred = True if inferred is None else False
-    name = 'emmo-inferred' if inferred else 'emmo'
+    name = 'emmo-inferred' if inferred in [True, None] else 'emmo'
     return get_ontology(name).load()
--- a/emmopy/emmopy.py
+++ b/emmopy/emmopy.py
@@ -1,8 +1,11 @@
 from ontopy import get_ontology
-from typing import Optional
+from typing import Optional, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from ontopy.ontology import Ontology
 
 
-def emmo(inferred: Optional[bool] = True):
+def get_emmo(inferred: Optional[bool] = True) -> "Ontology":
     """Returns the current version of emmo.
 
     Args:
@@ -10,10 +13,7 @@ def emmo(inferred: Optional[bool] = True):
         Default is True.
 
     Returns:
-        the loaded emmo ontology.
+        The loaded emmo ontology.
     """
-    if inferred == True:
-        name = 'emmo-inferred'
-    else:
-        name = 'emmo'
+    name = 'emmo-inferred' if inferred else 'emmo'
     return get_ontology(name).load()
--- a/emmopy/emmopy.py
+++ b/emmopy/emmopy.py
@@ -15,5 +15,6 @@ def get_emmo(inferred: Optional[bool] = True) -> "Ontology":
     Returns:
         The loaded emmo ontology.
     """
+    inferred = True if inferred is None else False
     name = 'emmo-inferred' if inferred else 'emmo'
     return get_ontology(name).load()
--- a/tests/test_load_emmo.py
+++ b/tests/test_load_emmo.py
@@ -6,10 +6,10 @@ def test_load_emmo() -> None:
     thisdir = os.path.abspath(os.path.dirname(__file__))
     sys.path.insert(1, os.path.abspath(os.path.join(thisdir, '..')))
     from ontopy import get_ontology  # noqa: E402, F401
-    from emmopy import emmo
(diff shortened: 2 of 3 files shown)


emmopy/emmopy.py
```diff
@@ -0,0 +1,19 @@
+from ontopy import get_ontology
+from typing import Optional
+
+
+def emmo(inferred: Optional[bool] = True):
```
CasperWA, line 5:
```suggestion
def emmo(inferred: bool = True):
```
Since `Optional[bool]` is equal to `Union[None, bool]`, essentially just adding `None` as an option, for this particular argument it isn't relevant. Unless you want to imply that people can pass `None` to `inferred`?
This would make sense, I think, and then if `inferred` is `None` it should be set to `True` in the function, since this is the default.

emmopy/emmopy.py
```diff
@@ -0,0 +1,19 @@
...
+    """Returns the current version of emmo.
```
CasperWA, line 6:
```suggestion
from typing import Optional, TYPE_CHECKING

if TYPE_CHECKING:
    from ontopy.ontology import Ontology


def emmo(inferred: Optional[bool] = True) -> "Ontology":
    """Returns the current version of emmo.
```

emmopy/emmopy.py
```diff
@@ -0,0 +1,19 @@
...
+
+    Args:
+        inferred: Whether to import the inferred version of emmo or not.
+        Default is True.
+
+    Returns:
+        the loaded emmo ontology.
```
CasperWA, line 13:
```suggestion
        The loaded emmo ontology.
```

emmopy/emmopy.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+    if inferred == True:
+        name = 'emmo-inferred'
+    else:
+        name = 'emmo'
```
CasperWA, line 18: name = 'emmo-inferred' if inferred else 'emmo'

tests/test_load_emmo.py
```diff
@@ -6,7 +6,10 @@ def test_load_emmo() -> None:
...
     from ontopy import get_ontology  # noqa: E402, F401
+    from emmopy import emmo
 
+    onto = get_ontology()
+    onto.load()
 
-    emmo = get_ontology()
-    emmo.load()
+    emmo = emmo()
+    emmo_inferred = emmo(inferred=False)
```
CasperWA, line 15: You're overwriting the variable for the function `emmo` with the returned value from said function here.

CasperWA (changes requested):
This is quite nice.
My comments are mainly for some typing tricks and such, not really the content.

Another thing I was thinking though was whether it should be named in capitals, where possible, instead of the lower case letters?

CasperWA:
> We could call it get_emmo instead.

Not a bad idea, it's still a bit confusing I think.
Mainly I was thinking to have something like:

```python
from emmopy import EMMO
```

Which would mean we could simply create an `EMMO` variable in the `__init__.py` file, or like you've done here, in a separate file (which is a bit nicer), and then import it in `__init__.py`. The trouble being that one will _always_ instantiate the EMMO ontology then when importing anything from `emmopy`. Unless it's done with some lazy loading tricks.

What I'm trying to say is, it might be best to rename to `get_emmo()` and then let users use this to create their own `EMMO` variables, like:

```python
from emmopy import get_emmo

EMMO = get_emmo(inferred=False)
EMMO_inferred = get_emmo()
```

emmopy/emmopy.py
francescalb: I was thinking inferred=None as a possible (although not common choice). I can remove it if you think that is just cluttering.

francescalb:
> > We could call it get_emmo instead.
> 
> Not a bad idea, it's still a bit confusing I think.
> Mainly I was thinking to have something like:
> 
> ```python
> from emmopy import EMMO
> ```
> 
> Which would mean we could simply create an `EMMO` variable in the `__init__.py` file, or like you've done here, in a separate file (which is a bit nicer), and then import it in `__init__.py`. The trouble being that one will _always_ instantiate the EMMO ontology then when importing anything from `emmopy`. Unless it's done with some lazy loading tricks.
> 
> What I'm trying to say is, it might be best to rename to `get_emmo()` and then let users use this to create their own `EMMO` variables, like:
> 
> ```python
> from emmopy import get_emmo
> 
> EMMO = get_emmo(inferred=False)
> EMMO_inferred = get_emmo()
> ```

I think get_emmo is better, since using `from emmopy import EMMO` would increase the risk of overwriting with `EMMO = EMMO()` later.

emmopy/emmopy.py
```diff
@@ -0,0 +1,19 @@
...
+    """Returns the current version of emmo.
+
+    Args:
+        inferred: Whether to import the inferred version of emmo or not.
+        Default is True.
+
+    Returns:
+        The loaded emmo ontology.
+    """
+    name = 'emmo-inferred' if inferred else 'emmo'
```
CasperWA, line 18:
```suggestion
    inferred = True if inferred is None else False
    name = 'emmo-inferred' if inferred else 'emmo'
```

tests/test_load_emmo.py
```diff
@@ -6,7 +6,10 @@ def test_load_emmo() -> None:
...
     sys.path.insert(1, os.path.abspath(os.path.join(thisdir, '..')))
     from ontopy import get_ontology  # noqa: E402, F401
-
+    from emmopy import get_emmo
 
     emmo = get_ontology()
     emmo.load()
+
+    EMMO_inferred = get_emmo()
+    EMMO = get_emmo(inferred=False)
```
CasperWA, line 15:
```suggestion
    EMMO = get_emmo(inferred=False)
    assert EMMO_inferred == get_emmo(None)
```

CasperWA (changes requested):

CasperWA (approved):
Merge at will if/when tests pass.

CasperWA (approved):
Par excellence

PR #239, opened 2021-09-28
Ensure all produced files from tests are in a temp dir

### LINKED ISSUE #220: Tidy up unittests
PR #208 moved to proper pytest functions for the tests, however, it didn't tidy up the tests.

The tests should be revisited and considered whether they can be optimized.

Most importantly, the tests shouldn't produce a lot of "garbage" files; all files should be temporary if at all.
For *examples* one can/should add these to the `examples/` folder or `demo/`.

CasperWA:
# Description:

Closes #220.

Add pytest fixtures for loading EMMO, as this is done in several tests.
Add fixture for location to repo_dir, instantiating it only once per pytest session.
Add a yield-fixture for creating a temporary directory that lives for the lifetime of the test function.

This removes the possibility to check through the generated files from the tests, keeping the unit tests being just that: unit tests.
However, I guess the generated files should be tested more now?

In short, the unit tests here should never rely on human inspection, hence, they shouldn't produce any permanent files.

If examples are the motive for the tests (as it is explicitly stated for one of them) it should be added to the `examples/` folder and be part of the documentation.

## Type of change:

- [ ] Bug fix.
- [X] New feature.
- [ ] Documentation update.

## Checklist:

- [x] Does this close the issue?
- [x] Is the change limited to the issue?

## Comments:

This *partly* closes the issue, since the tests should still be further tidied.


francescalb:
A lot of changes have been made here, not only ensuring that produced files do not persist after running pytest. I think most of them are cosmetic, but can we talk through them so I can learn a little here?

tests/conftest.py
```diff
@@ -0,0 +1,35 @@
...
+    return Path(__file__).parent.parent.resolve()
+
+
+@pytest.fixture
+def emmo() -> "Ontology":
+    """Load and return EMMO."""
+    from ontopy import get_ontology
+
+    emmo = get_ontology()
+    emmo.load()
```
francescalb, line 23:
```suggestion
    from emmopy import get_emmo

    emmo = get_emmo()
```

francescalb (changes requested):
I suggest loading emmo with get_emmo()

francescalb (approved):
Good

PR #240, opened 2021-09-28
Import from `collections.abc` when possible

### LINKED ISSUE #236: "Warning" Importing from `collections`
Instead of importing ABCs from `collections`, they should be imported from `collections.abc`. This is a general warning that pops up:

```console
EMMO-python/ontopy/factpluspluswrapper/sync_factpp.py:42: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working
    elif isinstance(ontology_or_world, collections.Sequence):
```

CasperWA:
# Description:

Fixes #236.

Importing ABCs from collections is deprecated and will be removed in Python 3.10.

As a note, I ran `pytest` and checked the resulting Python warnings. The only ones left are `DeprecationWarning`s the package raises itself.

## Type of change:

- [X] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

## Checklist:

### DIFF
--- a/demo/horizontal/emmo2meta.py
+++ b/demo/horizontal/emmo2meta.py
@@ -29,7 +29,7 @@ import owlready2
 if sys.version_info >= (3, 7):
     odict = dict
 else:
-    from collections.abc import OrderedDict as odict  # noqa: F401
+    from collections import OrderedDict as odict  # noqa: F401
 
 
 class EMMO2Meta:
--- a/demo/horizontal/emmo2meta.py
+++ b/demo/horizontal/emmo2meta.py
@@ -29,7 +29,7 @@ import owlready2
 if sys.version_info >= (3, 7):
     odict = dict
 else:
-    from collections import OrderedDict as odict  # noqa: F401
+    from collections.abc import OrderedDict as odict  # noqa: F401
 
 
 class EMMO2Meta:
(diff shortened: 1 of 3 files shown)


demo/horizontal/emmo2meta.py
```diff
@@ -29,7 +29,7 @@
 if sys.version_info >= (3, 7):
     odict = dict
 else:
-    from collections import OrderedDict as odict  # noqa: F401
+    from collections.abc import OrderedDict as odict  # noqa: F401
```
francescalb, line 32: I do not think OrderedDict is an abstract class? 

From the documentation here it is in collections.abc: https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes

However, the documentation [here](https://docs.python.org/3/library/collections.html) says that it will be moved. If so, that should apply for defaultdict as well. 

In my python3.8 importing OrderedDict from collections.abc does not work.

CasperWA: Indeed, you're right. Also, as this will only be imported in Python versions lower than 3.7, and this package only supports 3.6 and 3.7, this is only relevant for 3.6, which indeed imports from `collections` straight up 👍

francescalb (approved):
Perfect.

PR #245, opened 2021-09-29
Implement `pre-commit` & various tools

CasperWA:
# Description:

Closes #243.
~Partly addresses #223, but does not fully close it.~ Now fully closes #223.

Implement `pre-commit`.
Pre-commit is a tool that one installs in the git repository and runs prior to committing, checking various parts of the changed files.
For more information go to [pre-commit.com](https://pre-commit.com).

This slowly introduces `pre-commit` hooks.
First several hooks from the pre-commit-hooks repository maintained by pre-commit themselves are added:

- check-symlinks
- check-xml (running for all xml, rdf, and ttl files)
- check-yaml
- destroyed-symlinks
- end-of-file-fixer
- requirements-txt-fixer
- trailing-whitespace

To understand what each of these hooks do, go the [README for the pre-commit-hooks repo](https://github.com/pre-commit/pre-commit-hooks#readme).

Then `bandit` and `safety` will be introduced.
`bandit` will be introduced both as a `pre-commit` hooks, as well as running in the CI.
`safety` will only run in the CI.

## Type of change:

- [ ] Bug fix.
- [X] New feature.
- [ ] Documentation update.

## Checklist:

This checklist can be used as a help for the reviewer.

- [ ] Is the code easy to read and understand?
- [ ] Are comments for humans to read, not computers to disregard?
- [ ] Does a new feature has an accompanying new test (in the CI or unit testing schemes)?
- [ ] Has the documentation been updated as necessary?
- [ ] Does this close the issue?
- [ ] Is the change limited to the issue?
- [ ] Are errors handled for all outcomes?
- [ ] Does the new feature provide new restrictions on dependencies, and if so is this documented?

## Comments:

To do:
- [x] Setup pre-commit
- [X] Setup pre-commit-hooks hooks
- [X] Setup documentation build hooks (from `invoke` tasks)
- [X] Add `bandit`
- [X] Add `safety`
- [x] Add `pylint` (removing `flake8`)
- [x] Add `black`

PR #246, opened 2021-09-29
Added Standard methods to Ontology

### LINKED ISSUE #228: Standard dunder/magic methods for `Ontology`
It might be nice to implement some standard dunder/magic methods for the `Ontology` class facilitating standard logical operations on `Ontology` objects. This is at least true for equality (`==`).

This issue comes from the fact that we (@francescalb and I) wanted to test whether two instantiations of the `Ontology` class from the same file were equal or not. ([Here](https://github.com/emmo-repo/EMMO-python/runs/3699709937?check_suite_focus=true#step:6:41) is a reference to the failed test.)

It might be good to implement other dunder/magic methods as well, maybe even for other classes as well?
For more information on the special methods, see the [Python documentation](https://docs.python.org/3/reference/datamodel.html#special-method-names) or perhaps [this guide](https://rszalski.github.io/magicmethods/) written in 2012.

francescalb:
# Description:

Closes #228.

__hash__ (based on base_iri) defined for ontology.Ontology (only req to allow for __eq__ implementation)

__eq__  (based on all triples) defined for ontology.Ontology. Note that consistency between blank nodes is not checked.

This is tested in test_load_emmo.py.
The test also checks that two equal emmo-ontologies are no longer equal when a new class  or new relation is added to one of them.

## Type of change:

- [ ] Bug fix.
- [x] New feature.
- [ ] Documentation update.

## Checklist:

### DIFF
--- a/ontopy/ontology.py
+++ b/ontopy/ontology.py
@@ -211,10 +211,8 @@ class Ontology(owlready2.Ontology, OntoGraph):
             if isinstance(i, int):
                 if i >= 0:
                     return self._unabbreviate(i)
-                else:
-                    return "_:"  # blank nodes are given random neg. storid
-            else:
-                return i
+                return "_:"  # blank nodes are given random neg. storid
+            return i
 
         for s, p, o in self.world.get_triples():
             s = _unabbreviate(s)
--- a/ontopy/ontology.py
+++ b/ontopy/ontology.py
@@ -216,7 +216,9 @@ class Ontology(owlready2.Ontology, OntoGraph):
             return i
 
         for subject, predicate, obj in self.world.get_triples():
-            yield _unabbreviate(subject), _unabbreviate(predicate), _unabbreviate(obj)
+            yield (_unabbreviate(subject),
+                   _unabbreviate(predicate),
+                   _unabbreviate(obj))
 
     def get_by_label(self, label, label_annotations=None, namespace=None):
         """Returns entity with label annotation `label`.
--- a/ontopy/ontology.py
+++ b/ontopy/ontology.py
@@ -190,6 +190,26 @@ class Ontology(owlready2.Ontology, OntoGraph):
         # Play nice with inspect...
         pass
 
+    def __eq__(self, other):
+        return self.__hash__() == other.__hash__()
+
+    def __str__(self):
+        """TODO"""
+
+    def __iter__(self):
+        """TODO"""
+
+    def __delitem__(self):
+        """TODO"""
+
+    def __setitem__(self, key, value):
+        """ MAYBE TODO """
+        # self[key] = value
+        # emmo['HAtom'] = emmo.Atom
+
+    def __hash__(self):
+        return hash((self.base_iri, self.get_entities()))
+
     def get_by_label(self, label, label_annotations=None, namespace=None):
         """Returns entity with label annotation `label`.
 
--- a/ontopy/ontology.py
+++ b/ontopy/ontology.py
@@ -215,11 +215,8 @@ class Ontology(owlready2.Ontology, OntoGraph):
                 return "_:"  # blank nodes are given random neg. storid
             return i
 
-        for s, p, o in self.world.get_triples():
-            s = _unabbreviate(s)
-            p = _unabbreviate(p)
-            o = _unabbreviate(o)
-            yield s, p, o
+        for subject, predicate, obj in self.world.get_triples():
+            yield _unabbreviate(subject), _unabbreviate(predicate), _unabbreviate(obj)
 
     def get_by_label(self, label, label_annotations=None, namespace=None):
         """Returns entity with label annotation `label`.
--- a/ontopy/ontology.py
+++ b/ontopy/ontology.py
@@ -220,8 +220,8 @@ class Ontology(owlready2.Ontology, OntoGraph):
             self.get_entities(imported=True,
                               classes=True,
                               individuals=False,
-                              object_properties=False,
-                              data_properties=False,
+                              object_properties=True,
+                              data_properties=True,
                               annotation_properties=False,
                               ), key=lambda e: e.iri)
         return hash((self.base_iri, tuple(sorted_entities)))
--- a/ontopy/ontology.py
+++ b/ontopy/ontology.py
@@ -112,9 +112,6 @@ class World(owlready2.World):
 
         return onto
 
-    def __hash__(self):
-        return hash(self.filename)
-
 
 class Ontology(owlready2.Ontology, OntoGraph):
     """A generic class extending owlready2.Ontology.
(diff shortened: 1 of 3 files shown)


ontopy/ontology.py
```diff
@@ -190,6 +190,46 @@ def __objclass__(self):
...
+
+    def __eq__(self, other):
+        """Checks if this ontology is equal to other.
+
+        Equality of all triples obtained from self.get_triples(),
+        i.e. all triples except blank nodes.
+        """
+        return set(self.get_triples()) == set(other.get_triples())
+
+    def _sorted_entities(self):
```
CasperWA, line 207: When is this method used? I.e., what's the purpose of adding it in this PR?

ontopy/ontology.py
```diff
@@ -190,6 +190,46 @@ def __objclass__(self):
...
+        """Return sorted entities
+        """
+        return sorted(e.iri for e in self.get_entities(
```
CasperWA, line 210:
```suggestion
        return sorted(_.iri for _ in self.get_entities(
```
A trick in Python is to use an underscore (`_`) for an irrelevant variable. Here the `e` variable is, first of all, a bad name for a variable (it's difficult to understand what is meant by "e"), but other than that, it's not used for anything other than being the iterable variable. Mainly for the latter reason, you can use `_` instead, as it's not a necessary variable for further use.

ontopy/ontology.py
```diff
@@ -190,6 +190,46 @@ def __objclass__(self):
...
+        """ Returns all triples unabbreviated
+        """
+        def _unabbreviate(i):
+            if isinstance(i, int):
+                if i >= 0:
+                    return self._unabbreviate(i)
+                else:
+                    return "_:"  # blank nodes are given random neg. storid
+            else:
+                return i
```
CasperWA, line 225:
```suggestion
            return i
```

ontopy/ontology.py
```diff
@@ -190,6 +190,46 @@ def __objclass__(self):
...
+
+    def get_triples(self):
+        """ Returns all triples unabbreviated
+        """
+        def _unabbreviate(i):
+            if isinstance(i, int):
+                if i >= 0:
+                    return self._unabbreviate(i)
+                else:
+                    return "_:"  # blank nodes are given random neg. storid
```
CasperWA, line 223:
```suggestion
                return "_:"  # blank nodes are given random neg. storid
```

tests/test_load_emmo.py
```diff
@@ -7,10 +7,19 @@ def test_load_emmo() -> None:
     from emmopy import get_emmo
 
     EMMO_inferred = get_emmo()
+    EMMO_inf = get_emmo(None)
```
CasperWA, line 10:
```suggestion
    EMMO_inferred_again = get_emmo(None)
```

tests/test_load_emmo.py
```diff
@@ -7,10 +7,19 @@ def test_load_emmo() -> None:
...
     EMMO = get_emmo(inferred=False)
 
     assert EMMO_inferred
     assert EMMO
 
-    assert EMMO_inferred.base_iri == get_emmo(None).base_iri
+    assert EMMO_inferred == EMMO_inf
```
CasperWA, line 16:
```suggestion
    assert EMMO_inferred == EMMO_inferred_again
```

tests/test_load_emmo.py
```diff
@@ -7,10 +7,19 @@ def test_load_emmo() -> None:
...
+    assert EMMO != EMMO_inferred
 
+    EMMO_inferred.new_entity('HydrogenAtom', EMMO_inferred.Atom)
+    EMMO_inferred.sync_attributes()
+    assert EMMO_inferred != EMMO_inf
```
CasperWA, line 21:
```suggestion
    assert EMMO_inferred != EMMO_inferred_again
```

tests/test_load_emmo.py
```diff
@@ -7,10 +7,19 @@ def test_load_emmo() -> None:
...
+
+    EMMO_inferred2 = get_emmo()
+    EMMO_inferred2.Atom.comment.append('New triple')
+    assert EMMO_inferred2 != EMMO_inf
```
CasperWA, line 25:
```suggestion
    assert EMMO_inferred2 != EMMO_inferred_again
```

tests/test_load_emmo.py
```diff
@@ -7,10 +7,19 @@ def test_load_emmo() -> None:
     from emmopy import get_emmo
 
     EMMO_inferred = get_emmo()
+    EMMO_inf = get_emmo(None)
     EMMO = get_emmo(inferred=False)
 
     assert EMMO_inferred
     assert EMMO
```
CasperWA, line 14:
```suggestion
    assert EMMO_inferred_again
    assert EMMO
```

CasperWA (changes requested):
Looking great. I have only some minor code optimizations and a more significant question concerning the `_sorted_entities()` method.

ontopy/ontology.py
```diff
@@ -190,6 +190,37 @@ def __objclass__(self):
...
+                if i >= 0:
+                    return self._unabbreviate(i)
+                return "_:"  # blank nodes are given random neg. storid
+            return i
+
+        for s, p, o in self.world.get_triples():
+            s = _unabbreviate(s)
+            p = _unabbreviate(p)
+            o = _unabbreviate(o)
+            yield s, p, o
```
CasperWA, line 222:
```suggestion
        for subject, predicate, obj in self.world.get_triples():
            yield _unabbreviate(subject), _unabbreviate(predicate), _unabbreviate(obj)
```

CasperWA (changes requested):

francescalb:
Why?  I guess changing s to subject, p to predicate and o to obj can be done, but spo is quite standard when looking at triples I think. 

Then, doing the _unabreviate in the yield, do wish for that because it is faster or just because you think it is prettier? If it is faster I agree, but otherwise I am not a super fan of compacting code all the time, as it is more difficult to read for new people coming in I think.

francescalb:
OK, fixed.

jesper-friis:
Took a look at the conversation above. I am not sure I fully agree that using one-letter is always bad practice. In general I agree that descriptive variable names are good, but sensible use one-letter variables makes the the lines shorter and the code easier to read. If you just need an integer variable in a loop, I think that `i` or `n` are completely fine to use. The same is the case with `s`, `p` and `o` in triples. In contrary I am not much fan of `_` as a variable you actually use, but if you only want to loop over the objects in a set of triples I think it is fine to write `for _, _, o in triples: ...`.

CasperWA:
> Took a look at the conversation above. I am not sure I fully agree that using one-letter is always bad practice. In general I agree that descriptive variable names are good, but sensible use one-letter variables makes the the lines shorter and the code easier to read. If you just need an integer variable in a loop, I think that `i` or `n` are completely fine to use. The same is the case with `s`, `p` and `o` in triples. In contrary I am not much fan of `_` as a variable you actually use, but if you only want to loop over the objects in a set of triples I think it is fine to write `for _, _, o in triples: ...`.

The underscore is the accepted Pythonic way of defining un-used variables, specifically in the example you show above, but also in other cases, e.g., when a function returns a Triple type and all you need are one of those.

I think exactly because they are single-character variables they are _not_ easy to read, quite the contrary, it makes it impenetrable to understand, especially when the method or function names do not mention triples at all.
To align the variable naming a bit more, one can write out the "object" variable name, but append an underscore, as this is the way to avoid renaming built-ins according to PEP-8, i.e.: `for subject, predicate, object_ in get_me_some_triples(): ...`.

I agree that single-character variables can be used as iterating variables. This is common practice across several programming languages - but for ease of readability I'd like to strongly support more descriptive variable names throughout the code base.

CasperWA (approved):
This looks excellent to me now :) Thank you @francescalb !

CasperWA:
Everything said - we can always introduce our own standards for some variable naming, but I think it should be documented well then so there is a resource to go to if one is looking at this as a coder, but without an understanding of the content that's being worked with.

CasperWA:
This is all good for me. Let's discuss naming conventions perhaps in #245? As I'm doing a lot of different updates in the way of naming in that PR.

CasperWA:
Going through the code locally in my editor, I realize now you have overwritten the inherited `get_triples()` function. Was this intentional? If yes, perhaps it should reuse the same parameters and such? Either that, or the method should be renamed, e.g., to `get_unabbreviated_triples()` or something?

CasperWA:
The [latest commit from #245](https://github.com/emmo-repo/EMMO-python/pull/245/commits/5b67bf681df65982872ec8cf23af7a413e5a75da) includes my suggested change/fix for this.

francescalb:
You are right. That is not good at all. Your fix should be included.
Thanx!

CasperWA:
Great - I'll make an extra issue and PR for it and redact my other PR.

PR #283, opened 2021-11-06
Improved representation of blank nodes

jesper-friis:
# Description:
* Changed get_unabbriviated_triples() to represent blank nodes such they can be used for reasoning.
* Has not changed the result of `Ontology.__eq__()`.
* Added get_unabbriviated_triples() to World too, which will return all triples in the world.

## Type of change:
- [x] Bug fix.
- [x] New feature.
- [ ] Documentation update.

## Checklist:

### DIFF
--- a/ontopy/ontology.py
+++ b/ontopy/ontology.py
@@ -52,6 +52,27 @@ DEFAULT_LABEL_ANNOTATIONS = [
 ]
 
 
+class BNode:
+    """Represents a blank node."""
+    def __init__(self, base_iri, storid):
+        if (storid >= 0):
+            raise ValueError(
+                f'bnode is supposed to have a negative storid: {storid}')
+        self.base_iri = base_iri
+        self.storid = storid
+
+    def __repr__(self):
+        return repr(f'_:b{-self.storid}')
+
+    def __hash__(self):
+        return hash((self.base_iri, self.storid))
+
+    def __eq__(self, other):
+        """For now blank nodes always compare true with each other."""
+        print('***', self, other, isinstance(other, BNode))
+        return isinstance(other, BNode)
+
+
 def get_ontology(*args, **kwargs):
     """Returns a new Ontology from `base_iri`.
 
@@ -202,13 +223,13 @@ class Ontology(  # pylint: disable=too-many-public-methods
         pass
 
     def __hash__(self):
-        """Returns hash based on base_iri
+        """Returns a hash based on base_iri.
         This is done to keep Ontology hashable when defining __eq__.
         """
         return hash(self.base_iri)
 
     def __eq__(self, other):
-        """Checks if this ontology is equal to other.
+        """Checks if this ontology is equal to `other`.
 
         Equality of all triples obtained from self.get_unabbreviated_triples(),
         i.e. blank nodes are not distinguished, but relations
@@ -219,13 +240,15 @@ class Ontology(  # pylint: disable=too-many-public-methods
         )
 
     def get_unabbreviated_triples(self):
-        """Returns all triples unabbreviated"""
+        """Returns all triples unabbreviated."""
 
         def _unabbreviate(i):
             if isinstance(i, int):
                 if i >= 0:
                     return self._unabbreviate(i)
-                return "_:"  # blank nodes are given random neg. storid
+                # negative storid corresponds to blank nodes
+                #return '_:b'
+                return BNode(self.base_iri, i)
             return i
 
         for subject, predicate, obj in self.get_triples():
--- a/ontopy/ontology.py
+++ b/ontopy/ontology.py
@@ -28,9 +28,11 @@ from owlready2 import locstr
 from owlready2.entity import ThingClass
 
 from ontopy.factpluspluswrapper.sync_factpp import sync_reasoner_factpp
-
-from ontopy.utils import asstring, read_catalog, infer_version, convert_imported
 from ontopy.utils import (
+    asstring,
+    read_catalog,
+    infer_version,
+    convert_imported,
     FMAP,
     IncompatibleVersion,
     isinteractive,
@@ -116,26 +118,26 @@ class World(owlready2.World):
 
         return onto
 
-    def get_unabbreviated_triples(self, bnode=None):
-        """ Returns all triples unabbreviated.
+    def get_unabbreviated_triples(self, label=None):
+        """Returns all triples unabbreviated.
 
-        If `bnode` is given, it will be used to represent blank nodes.
+        If `label` is given, it will be used to represent blank nodes.
         """
+
         def _unabbreviate(i):
             if isinstance(i, int):
                 # negative storid corresponds to blank nodes
                 if i >= 0:
                     return self._unabbreviate(i)
-                elif bnode is None:
-                    return BNode(self, i)
-                else:
-                    return bnode
+                return BlankNode(self, i) if label is None else label
             return i
 
         for subject, predicate, obj in self.get_triples():
-            yield (_unabbreviate(subject),
-                   _unabbreviate(predicate),
-                   _unabbreviate(obj))
+            yield (
+                _unabbreviate(subject),
+                _unabbreviate(predicate),
+                _unabbreviate(obj),
+            )
 
 
 class Ontology(  # pylint: disable=too-many-public-methods
@@ -232,20 +234,20 @@ class Ontology(  # pylint: disable=too-many-public-methods
         """Checks if this ontology is equal to `other`.
 
         This function compares the result of
-        ``set(self.get_unabbreviated_triples(bnode='_:b'))``,
+        ``set(self.get_unabbreviated_triples(label='_:b'))``,
         i.e. blank nodes are not distinguished, but relations to blank
         nodes are included.
         """
-        return set(self.get_unabbreviated_triples(bnode='_:b')) == set(
-            other.get_unabbreviated_triples(bnode='_:b')
+        return set(self.get_unabbreviated_triples(label="_:b")) == set(
+            other.get_unabbreviated_triples(label="_:b")
         )
 
-    def get_unabbreviated_triples(self, bnode=None):
-        """ Returns all triples unabbreviated.
+    def get_unabbreviated_triples(self, label=None):
+        """Returns all triples unabbreviated.
 
-        If `bnode` is given, it will be used to represent blank nodes.
+        If `label` is given, it will be used to represent blank nodes.
         """
-        return World.get_unabbreviated_triples(self, bnode)
+        return World.get_unabbreviated_triples(self, label)
 
     def get_by_label(
         self, label, label_annotations=None, namespace=None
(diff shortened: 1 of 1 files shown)


francescalb (approved):
Looks nice. Smart to put the get_unabbreviated_triples in world.

ontopy/ontology.py
```diff
@@ -1089,3 +1104,31 @@ def new_entity(self, name, parent):
...
+            raise ValueError(
+                f'bnode is supposed to have a negative storid: {storid}')
+        self.onto = onto
+        self.storid = storid
+
+    def __repr__(self):
+        return repr(f'_:b{-self.storid}')
+
+    def __hash__(self):
+        return hash((self.onto, self.storid))
```
CasperWA, line 1130:
```suggestion
        return hash(tuple(getattr(self, _) for _ in sorted(self.__dict__))
```
To make sure all attributes (and future attributes) are always given to the hash-function in an alphabetically sorted tuple of attribute values.

ontopy/ontology.py
```diff
@@ -191,36 +213,29 @@ def __objclass__(self):
...
+    def get_unabbreviated_triples(self, bnode=None):
+        """ Returns all triples unabbreviated.
 
-        for subject, predicate, obj in self.get_triples():
-            yield (_unabbreviate(subject),
-                   _unabbreviate(predicate),
-                   _unabbreviate(obj))
+        If `bnode` is given, it will be used to represent blank nodes.
+        """
+        return World.get_unabbreviated_triples(self, bnode)
```
CasperWA, line 238:
```suggestion
        return self.world.get_unabbreviated_triples(self, bnode)
```
Shouldn't this be using the method from it's stored world ontology, or?

ontopy/ontology.py
```diff
@@ -191,36 +213,29 @@ def __objclass__(self):
...
-        """ Returns all triples unabbreviated
-        """
-        def _unabbreviate(i):
-            if isinstance(i, int):
-                if i >= 0:
-                    return self._unabbreviate(i)
-                return "_:"  # blank nodes are given random neg. storid
-            return i
+    def get_unabbreviated_triples(self, bnode=None):
+        """ Returns all triples unabbreviated.
```
CasperWA, line 234:
```suggestion
        """Returns all triples unabbreviated.
```

ontopy/ontology.py
```diff
@@ -191,36 +213,29 @@ def __objclass__(self):
...
-    def get_unabbreviated_triples(self):
-        """ Returns all triples unabbreviated
-        """
-        def _unabbreviate(i):
-            if isinstance(i, int):
-                if i >= 0:
-                    return self._unabbreviate(i)
-                return "_:"  # blank nodes are given random neg. storid
-            return i
+    def get_unabbreviated_triples(self, bnode=None):
```
CasperWA, line 233:
```suggestion
    def get_unabbreviated_triples(self, blank_node=None):
```

ontopy/ontology.py
```diff
@@ -191,36 +213,29 @@ def __objclass__(self):
...
+        """ Returns all triples unabbreviated.
 
-        for subject, predicate, obj in self.get_triples():
-            yield (_unabbreviate(subject),
-                   _unabbreviate(predicate),
-                   _unabbreviate(obj))
+        If `bnode` is given, it will be used to represent blank nodes.
```
CasperWA, line 236:
```suggestion
        If `blank_node` is given, it will be used to represent blank nodes.
```

ontopy/ontology.py
```diff
@@ -191,36 +213,29 @@ def __objclass__(self):
...
 
     def __eq__(self, other):
-        """Checks if this ontology is equal to other.
+        """Checks if this ontology is equal to `other`.
 
-        Equality of all triples obtained from self.get_unabbreviated_triples(),
-        i.e. blank nodes are not distinguished, but relations
-        to blank nodes are included.
+        This function compares the result of
+        ``set(self.get_unabbreviated_triples(bnode='_:b'))``,
```
CasperWA, line 225:
```suggestion
        `set(self.get_unabbreviated_triples(bnode='_:b'))`,
```

ontopy/ontology.py
```diff
@@ -112,6 +113,27 @@ def get_ontology(self, base_iri='emmo-inferred'):
 
         return onto
 
+    def get_unabbreviated_triples(self, bnode=None):
```
CasperWA, line 116:
```suggestion
    def get_unabbreviated_triples(self, blank_node=None):
```

ontopy/ontology.py
```diff
@@ -112,6 +113,27 @@ def get_ontology(self, base_iri='emmo-inferred'):
...
+        """ Returns all triples unabbreviated.
```
CasperWA, line 117:
```suggestion
        """Returns all triples unabbreviated.
```

ontopy/ontology.py
```diff
@@ -112,6 +113,27 @@ def get_ontology(self, base_iri='emmo-inferred'):
...
+
+        If `bnode` is given, it will be used to represent blank nodes.
```
CasperWA, line 119:
```suggestion
        If `blank_node` is given, it will be used to represent blank nodes.
```

ontopy/ontology.py
```diff
@@ -112,6 +113,27 @@ def get_ontology(self, base_iri='emmo-inferred'):
...
+        """
+        def _unabbreviate(i):
+            if isinstance(i, int):
+                # negative storid corresponds to blank nodes
+                if i >= 0:
+                    return self._unabbreviate(i)
+                elif bnode is None:
+                    return BNode(self, i)
```
CasperWA, line 127:
```suggestion
                    return BlankNode(self, i)
```

ontopy/ontology.py
```diff
@@ -1089,3 +1104,31 @@ def new_entity(self, name, parent):
...
+
+    def __repr__(self):
+        return repr(f'_:b{-self.storid}')
+
+    def __hash__(self):
+        return hash((self.onto, self.storid))
+
+    def __eq__(self, other):
+        """For now blank nodes always compare true against each other."""
+        return isinstance(other, BNode)
```
CasperWA, line 1134:
```suggestion
        return isinstance(other, BlankNode)
```

CasperWA (changes requested):
Cheers @jesper-friis.

I'd like to propose writing out BNode to BlankNode to not confuse with "binary" or something else.

I've tried to add a bit more description to it as well, feel free to adapt it as you please.

One final thing: It's quite unclear to me what `blank_node` can be in the `get_unabbreviated_triples()` methods. Should it be a string (as you do in `__eq__()` or should it be a `BlankNode` type? I'm pretty sure some of the code only works with one or the other - specifically the internal `_unabbreviate()` function.

ontopy/ontology.py
CasperWA: Fixed by `black`.

ontopy/ontology.py
CasperWA: Fixed by `black`.

ontopy/ontology.py
```diff
@@ -1089,3 +1104,31 @@ def new_entity(self, name, parent):
...
+    def __init__(self,
+                 onto: Union[World, Ontology],
+                 storid: int):
+        """Initiate a blank node.
+
+        Args:
+            onto: Ontology or World instance.
+            storid: The storage id of the blank node.
+        """
+        if (storid >= 0):
```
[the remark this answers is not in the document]
CasperWA: Fixed by `black`.

ontopy/ontology.py
```diff
@@ -1217,3 +1232,31 @@ def new_entity(self, name, parent):
...
+    def __init__(self, onto: Union[World, Ontology], storid: int):
+        """Initiate a blank node.
+
+        Args:
+            onto: Ontology or World instance.
+            storid: The storage id of the blank node.
+        """
+        if storid >= 0:
+            raise ValueError(
+                f"bnode is supposed to have a negative storid: {storid}"
```
CasperWA, line 1249:
```suggestion
                f"A BlankNode is supposed to have a negative storid: {storid}"
```

ontopy/ontology.py
```diff
@@ -1217,3 +1232,31 @@ def new_entity(self, name, parent):
...
+            )
+        self.onto = onto
+        self.storid = storid
+
+    def __repr__(self):
+        return repr(f"_:b{-self.storid}")
```
CasperWA, line 1255: So, here you're representing it with the positive integer. I.e., a blank node with `storid` of `-5` becomes `"_:b5"`?

CasperWA (changes requested):
Re-adding the now outdated comments from my recent review.

ontopy/ontology.py
```diff
@@ -1089,3 +1104,31 @@ def new_entity(self, name, parent):
         with self:
             e = types.new_class(name, (parent, ))
         return e
+
+
+class BNode:
```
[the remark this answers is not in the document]
jesper-friis: I think it should be fine to use "bnode", it is a commonly used term within RDF, see eg. https://en.wikipedia.org/wiki/bnode

ontopy/ontology.py
```diff
@@ -1089,3 +1104,31 @@ def new_entity(self, name, parent):
...
+    """Represents a blank node."""
```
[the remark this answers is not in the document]
jesper-friis: A bnode is a node, not a triple. I have pushed an updated description.

ontopy/ontology.py
jesper-friis: Yes, that is intentional, "_:b-5" would be ugly and could cause problems for some tools. 

bnodes have no public IRI, so it doesn't matter how we choose to represent as long we are internally consistent.

jesper-friis:
> One final thing: It's quite unclear to me what `blank_node` can be in the `get_unabbreviated_triples()` methods. Should it be a string (as you do in `__eq__()` or should it be a `BlankNode` type? I'm pretty sure some of the code only works with one or the other - specifically the internal `_unabbreviate()` function.

The idea is to represent blank nodes as BNode objects. However, since bnodes are local, it becomes tricky to compare them across imported ontologies. The purpose of introducing the `bnode` argument of `_unabbriveate()` is to allow `__eq__()` to represent all bnodes in a similar manner. Anything could have been used, but a string is just simple and fast. Since it is internal to the `__eq__()` function, it doesn't affect anything outside it.

ontopy/ontology.py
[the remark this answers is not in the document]
CasperWA: The fact that it redirects to "Blank node" is also a very fair reason to change from `BNode` to `BlankNode` in my opinion.
image
But in the end one is simply an alias for the other. If you strongly prefer `BNode` that's fine. This was just a suggestion to make the code more readable for the uninitiated :)

ontopy/ontology.py
```diff
@@ -191,36 +213,29 @@ def __objclass__(self):
         pass
 
     def __hash__(self):
-        """ Returns hash based on base_iri
+        """Returns a hash based on base_iri.
```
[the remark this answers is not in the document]
CasperWA: I don't see that you have used this suggestion, but you have nonetheless marked this conversation as resolved - you don't think this change makes sense or you just haven't pushed the local changes?

francescalb:
@CasperWA @jesper-friis Please decide on whether we should have bnode or blank_node as variable, just make sure that we are consistent. I do not really have very strong preferences. I agree that blank_node is more understandable, but I do not like to long variable names as I find codes more difficult to read when they become long. The reason the tests fails now is because bnode is only changed to blank_node in one place.

jesper-friis:
I tried to accept Casper's suggestions about renaming bnode to blank_node. The last inconsistencies are now removed. @CasperWA, are you happy to merge now? 

I kept `bnode` as the argument name of get_unabbreviated_triples() since it not intended to be an instance of BlankNode. Probably something as generic as `s` would be an even better naming of the argument.

CasperWA:
> I kept `bnode` as the argument name of get_unabbreviated_triples() since it not intended to be an instance of BlankNode. Probably something as generic as `s` would be an even better naming of the argument.

Yes, please update this to `subject`, `entity` or similar. Then I think this should be good to go. Let me just do a proper review after this change first, though.

jesper-friis:
Done. Changed "label"

CasperWA (approved):
Thanks @jesper-friis - I'll merge this in now.

PR #291, opened 2021-11-10
Added test for label name length in ontology.new_entity

francescalb:
# Description:

Added check that the label of a new entity consists of only one word in ontology.new_entity.

## Type of change:

- [ ] Bug fix.
- [x] New feature.
- [ ] Documentation update.

## Checklist:


ontopy/ontology.py
```diff
@@ -1208,12 +1210,19 @@ def get_wu_palmer_measure(self, cls1, cls2):
...
 
         Makes a new entity in the ontology with given parent.
-        Return the new entity
+        Return the new entity.
+
+        Throws exception if name consists of more than one word.
         """
+        if len(name.split(" ")) > 1:
+            raise LabelDefinitionError(
+                f"Error in label name definition {name}"
```
CasperWA, line 1223:
```suggestion
                f"Error in label name definition {name}: "
```

ontopy/utils.py
```diff
@@ -50,6 +50,10 @@ class NoSuchLabelError(LookupError, AttributeError):
     """Error raised when a label cannot be found."""
 
 
+class LabelDefinitionError(Exception):
+    """Error in label definition"""
```
CasperWA, line 54:
```suggestion
    """Error in label definition."""
```

CasperWA (approved):
Very nice!

Are all entities `ThingClass`? Great to know. This should be the default type for entities throughout the code in the future then. Or do we have a more EMMOntoPy-specific class?

Concerning the two suggested changes only the one for the error message is important.
And no need to include me as contributing author for these changes. I suggest just squash-merging and removing my name from the commit message 👍

PR #295, opened 2021-11-17
Added functionality for more than one parent in new_entity

francescalb:
# Description:

ontology.new_entity now accepts a single parent or a list of parents.

## Type of change:

- [ ] Bug fix.
- [x] New feature.
- [ ] Documentation update.

## Checklist:

### DIFF
--- a/ontopy/ontology.py
+++ b/ontopy/ontology.py
@@ -1216,7 +1216,7 @@ class Ontology(  # pylint: disable=too-many-public-methods
     ) -> ThingClass:
         """Create and return new entity
 
-        Makes a new entity in the ontology with given parent.
+        Makes a new entity in the ontology with given parent(s).
         Return the new entity.
 
         Throws exception if name consists of more than one word.
--- a/ontopy/ontology.py
+++ b/ontopy/ontology.py
@@ -16,6 +16,7 @@ import warnings
 import uuid
 import tempfile
 import types
+from typing import Union, List
 from collections import defaultdict
 
 import rdflib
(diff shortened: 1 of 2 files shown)


ontopy/ontology.py
```diff
@@ -1223,6 +1226,8 @@ def new_entity(self, name: str, parent: ThingClass) -> ThingClass:
                 f"Error in label name definition {name}: "
                 "Label consists of more than one word."
             )
-        with self:
-            entity = types.new_class(name, (parent,))
+        parent = parent if isinstance(parent, list) else [parent]
```
jesper-friis, line 1229: Better to test against `collections.abc.Sequence` than just `list` to accept any list-like type, like tuples.

jesper-friis: Maybe you can call the new variable `parents` to emphasise that it is a sequence. Then you even avoid modifying the argument, which is considered good practice.

ontopy/ontology.py
```diff
@@ -1223,6 +1226,8 @@ def new_entity(self, name: str, parent: ThingClass) -> ThingClass:
...
+        for thing in parent:
+            with self:
+                entity = types.new_class(name, (thing,))
```
jesper-friis, line 1232: This will create one new entity for each parent, which is not what we want. Simply replace the for loop with

```python
with self:
```

francescalb: It will not, but I agree it is more elegant to do it in a oneliner, need to make parents a tuple.

ontopy/ontology.py
francescalb: testing against a collections.abc.Iterable now, and renamed to parents.

jesper-friis (approved):

Issue #299, opened 2021-11-23
Check out how to get version of ontology


francescalb:
Currently get_ontology of emmo-inferred (latest-stable) fails as versionIRI is missing. 
Look into how to make better error message and perhaps infer from other tags if versionIRI is missing.

Also ontoversion fails and must be checked.

francescalb:
There is already a sufficient error message in ontology.get_version(), which is:
TypeError: No versionIRI in Ontology 

Inferring from other tags I think is a bad idea as this will open up for mistakes that are difficult to track later.

The default version in ontoversion should be ttl, and it would be nice to have a better error message here, that suggests including format.

CasperWA:
Determine whether `versionInfo` or `versionIRI` should be the higher source of truth. And at least use both to try and determine the version.

francescalb:
@jesper-friis 

I want to add the possibility of getting the version from the versionInfo as well as from the versionIRI. Which one do you think should have precedence?

jesper-friis:
Since the semantics of owl:versionInfo is not well defined on https://www.w3.org/TR/owl-ref/#versionInfo-def, I think versionIRI should have precedence.

However, on https://www.w3.org/TR/2004/REC-owl-guide-20040210/#OntologyVersioning they indicate that `owl:versionInfo` can be used by versioning systems. So good practice is probably to set owl:versionInfo to the version number (without any prepended "v").

francescalb:
OK, from the above discussion I propose to do as follows:

if as_iri:
     return VersionIRI of versionIRI else Exception versionIRImissing
else:
    if version_iri:
        return version inferred from version_iri and base_iri
    elif versionInfo:
         return versionINFO
    else:
        raise NoVersionInfoFound....
    
    


I guess, then the only reason make this change is to find version from versionInfo if the ontology is malformed and does not have versionIRI (which it should have). Comments?

francescalb:
> 

@jesper-friis , you were quick on approving the draft, can you please comment on this?

PR #301, opened 2021-11-24
Check first versionIRI then versionInfo in ontology.get_version()

francescalb:
# Description:

Improve on get_version in ontology so that also versionInfo is checked for. 

See discussion on #299 

Also changed so that rdflib can guess format in ontoversion. 

Ontoversion only checks for version from versionIRI as this should be the case for a correctly defined ontology.

## Type of change:

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

## Checklist:


tools/ontoversion
```diff
@@ -8,21 +8,11 @@ import argparse
 import sys
 
 import rdflib
+from rdflib.exceptions import ParserError
```
jesper-friis, line 11: Not needed any more as far as I can see
```suggestion
```

francescalb: Yes, you are right. This is not working properly yet though so it is still only a draft pull request, and I am keeping some things in the code for help until when I can look at it again.  I'll let you know when review is needed again.

francescalb:
@jesper-friis 
Can you please look at the ontology.get_version() now and see if you think this is the way to go forward?
I am now giving precedence to versionInfo, but this should be discussed.

the tools/ontoversion is not implemented correctly yet as I want us to decide if this is the right approach first.

ontopy/ontology.py
```diff
@@ -1051,23 +1051,38 @@ def is_defined(self, entity):
...
-        version_iri = self.world._unabbreviate(obj)
-        if as_iri:
-            return version_iri
-        return infer_version(self.base_iri, version_iri)
+        if isinstance(obj, int):
+            version_iri = self.world._unabbreviate(obj)
+        else:
+            version_iri = obj.strip('"').strip("'")
+
+        if as_iri and not version_iri_storid:
```
jesper-friis, line 1081: The case `as_iri=True` and `version_iri_storid` is given is not considered. In that case we should simply return `versionIRI`. 

Note also the case where the ontology provides both `versionInfo` and `versionIRI`. If `as_iri` is true, we want to return `versionIRI`, but that is never fetched due to the if-statement on line 1065.

francescalb: if ´as_iri=True´ and ´version_iri_storid´ the last return statement return versionIRI, no?

if ´as_iri´ is true and version_info exists version_iri is constructed from base_iri. I agree this is not a good solution. This means that it should check for versionIRI alwasy is as_iri is true.

jesper-friis: Yes, you are right.

jesper-friis (approved):

jesper-friis (approved):
Looks very good. 

My only question is whether we should encourage setting versionInfo to just be the version number without the prepended "v".

francescalb:
> Looks very good.
> 
> My only question is whether we should encourage setting versionInfo to just be the version number without the prepended "v".

Sure, it did it to keep track of versionIRI vs versionInfo, but it should be removed now.

Issue #223, opened 2021-09-21
Update code styling and linting


CasperWA:
This is very much a developer's issue.

I'd like to suggest to update the Python code styling to use [`black`](https://black.readthedocs.io/en/stable/), and otherwise use [`pylint`](https://pylint.pycqa.org/en/latest/) for linting (instead of `flake8` as the CI job is doing now).
Concerning the linting I don't have as strong feelings as I do with respect to the styling/formatting, however, in my experience `pylint` seems more transparent to me with better checks.

Furthermore, I'd like to add other code-checking tools to this code, to make sure it's secure and developer-friendly.
Taken directly from @quuat 's [FastAPI template](https://github.com/SINTEF/fastapi-template), I'd want to implement the following tools (in addition to `black` and `pylint`):
- [`bandit`](https://bandit.readthedocs.io/en/latest/)
  Checks code security, e.g., when using URL calls, etc.
- [`safety`](https://github.com/pyupio/safety#readme)
  Checks all dependencies for known security vulnerabilities.
- [`mypy`](https://mypy.readthedocs.io/en/stable/)
  A static type checker.

The latter (`mypy`) might be "controversial", however, I think having a good typing allows for clearer code, since it's clear what a function will expect of inputs and what it will eventually return. Typing up the code also makes for better IDE experiences.

Personally, I would implement this as a combination between [`pre-commit`](https://pre-commit.com/) and CI (GitHub Action job).
The `pre-commit` tool is quite neat as it can auto-reformat the code according to our chosen code styling before we commit, as well as sanity check our new code before committing.
However, it would need to be installed in every working clone one has by doing `pre-commit install`, although only once.
Also, it can always (but should never) be circumvented by passing the `--no-verify` option to `git commit`.
These reasons are why it should be implemented also in the CI.

The reason for not just having it in the CI is that the obvious mistakes get caught before committing and running external resources. This is very good and makes for quicker development, when we don't have to wait for stupid mistakes or typos to first become apparent after the CI job has finished.

As a note, I recently implemented these tools in [a project of mine](https://github.com/Materials-Consortia/optimade-gateway/commit/6d3b9723f7f0759dfdc311a17d315fd629a97769).

CasperWA:
Should I go ahead with this? It will take a bit of work to create the PR, so I don't want to start it unless it's something we want done.

francescalb:
[...]

PR #307, opened 2021-11-30
Fix ontoconvert rdflib import

### LINKED ISSUE #306: Ontodoc failing due to wrong `rdflib` import
This issue was found during documentation production in BattINFO (see [this job](https://github.com/BIG-MAP/BattINFO/runs/4355533461?check_suite_focus=true#step:10:32)).

CasperWA:
# Description:

Fixes #306 

`Graph` should be imported from `rdflib.graph` in the current version of `rdflib`.

Added simple tests for ensuring the `main()` functions for each tool runs correctly.

## Type of change:

- [X] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

## Checklist:


emmopy/emmocheck.py
```diff
@@ -463,7 +463,9 @@ def checker(onto, ignore_namespace):
         checker(self.onto, self.ignore_namespace)
 
 
-def main():  # pylint: disable=too-many-locals,too-many-branches,too-many-statements
+def main(
+    argv: list = None,
```
jesper-friis, line 467: What is the purpose of this argument?

tests/tools/test_emmocheck.py
```diff
@@ -0,0 +1,14 @@
+"""Test the `emmocheck` tool."""
+import pytest
+
+
+@pytest.mark.parametrize("tool", ["emmocheck"], indirect=True)
+def test_run(tool) -> None:
+    """Check running `emmocheck` works."""
```
jesper-friis, line 7:
```suggestion
    """Check that running `emmocheck` works."""
```

tests/tools/conftest.py
```diff
@@ -0,0 +1,34 @@
+from typing import TYPE_CHECKING
```
jesper-friis, line 1: I am not smart enough to see what actually are testing here. Maybe you can add a comment

tests/tools/test_ontoconvert.py
```diff
@@ -0,0 +1,14 @@
+"""Test the `ontoconvert` tool."""
+from pathlib import Path
+
+import pytest
+
+
+@pytest.mark.parametrize("tool", ["ontoconvert"], indirect=True)
+def test_run(tool, tmpdir: Path) -> None:
+    """Check running `ontoconvert` works."""
```
jesper-friis, line 9:
```suggestion
    """Check that running `ontoconvert` works."""
```

tests/tools/test_ontoconvert.py
```diff
@@ -0,0 +1,14 @@
...
+    test_file = (
+        Path(__file__).resolve().parent.parent / "testonto" / "models.ttl"
+    )
+
+    tool.main([str(test_file), str(tmpdir / "test.ttl")])
```
jesper-friis, line 14: As far as I can see this only checks that `ontoconvert` runs. Shouldn't the result also be checked?

tests/tools/test_ontodoc.py
```diff
@@ -0,0 +1,15 @@
+"""Test the `ontodoc` tool."""
+from pathlib import Path
+
+import pytest
+
+
+@pytest.mark.skip("ontodoc is tested in other ways")
+@pytest.mark.parametrize("tool", ["ontodoc"], indirect=True)
+def test_run(tool, tmpdir: Path) -> None:
+    """Check running `ontodoc` works."""
```
jesper-friis, line 10:
```suggestion
    """Check that running `ontodoc` works."""
```

tests/tools/test_ontodoc.py
```diff
@@ -0,0 +1,15 @@
...
+    test_file = (
+        Path(__file__).resolve().parent.parent / "testonto" / "models.ttl"
+    )
+
+    tool.main([str(test_file), str(tmpdir / "test.ttl")])
```
jesper-friis, line 15: The output of ontodoc should be something supported by pandoc. 
```suggestion
    tool.main([str(test_file), str(tmpdir / "test.md")])
```

tests/tools/test_ontograph.py
```diff
@@ -0,0 +1,14 @@
+"""Test the `ontograph` tool."""
+from pathlib import Path
+
+import pytest
+
+
+@pytest.mark.parametrize("tool", ["ontograph"], indirect=True)
+def test_run(tool, tmpdir: Path) -> None:
+    """Check running `ontograph` works."""
```
jesper-friis, line 9:
```suggestion
    """Check that running `ontograph` works."""
```

tests/tools/test_ontograph.py
```diff
@@ -0,0 +1,14 @@
...
+    test_file = (
+        Path(__file__).resolve().parent.parent / "testonto" / "models.ttl"
+    )
+
+    tool.main([str(test_file), str(tmpdir / "test.ttl")])
```
jesper-friis, line 14:
```suggestion
    tool.main([str(test_file), str(tmpdir / "test.png")])
```

emmopy/emmocheck.py
jesper-friis: Ahh, I see it now why you introduced `argv`. It is used by the tests. 

It this the preferred way to test python tools - by calling their main() function from within Python?
Wouldn't it be more direct to use subprocess to check them by calling them as they would have been called from the shell?

jesper-friis: If we keep argv, I think it would be good with a comment to explain why it is there.

jesper-friis (commented):
Adds simple tests of all tools. Nice

tests/tools/conftest.py
```diff
@@ -0,0 +1,34 @@
...
+def tool(request: "Dict[str, Any]") -> "ModuleType":
+    """Import a tool as a module."""
+    from copy import deepcopy
+    import importlib
+    from pathlib import Path
+    import sys
+
+    original_sys_path = deepcopy(sys.path)
+    original_tool_path: Path = (
+        Path(__file__).resolve().parent.parent.parent / "tools" / request.param
```
francescalb, line 20: Why do we have to add the path specifically? Aren't the tools in sys.path when installed through emmontopy?

emmopy/emmocheck.py
CasperWA: Sure. And I'd say that one can test it in both ways. I would test it through Python to be able to keep everything in a "closed pytest environment", but at the same time, a `multiprocessing.Process` or `submodule.run` approach would be good to have as well to make sure the shell calls work as intended.
In other words - I wouldn't remove this work, but simply add to it.

And sure - I can create a comment in the `main()` functions' doc-strings.

tests/tools/conftest.py
CasperWA: This is for type annotations only. It has nothing to do with testing, but rather to support IDEs as well as set the package up for future testing with MyPy (see #223).

tests/tools/test_ontoconvert.py
CasperWA: No, the issue here was that running `ontoconvert` was failing - so here I've created extremely simple tests to make sure it can actually run as intended. Then we can make extra tests to check the contents and logic as well. I think modularizing tests this way makes it easier to manage the testing suite, as each test focuses on a single (simple) feature that it's checking. Hopefully that will make it easier to pinpoint bugs in the future as well :)

tests/tools/conftest.py
CasperWA: The tools edit `sys.path`, therefore, I'm storing and restoring `sys.path` to it's original value before and after a test (that uses this fixture).
We should reconsider the way this is done, to be honest, either relying on non-Python scripts to run the tools if it's not desirable to install the package and otherwise simply use `setuptools` to define the tools. This will auto-generate `sh` files in a place that will be added to the environment's `PATH` variable automatically. Adding it manually is --- user unfriendly :)

francescalb: Thank you for the explanation.

francescalb (approved):
Looks nice, good to have tests for the tools as well!

jesper-friis (approved):

tests/tools/conftest.py
jesper-friis: Thanks for the explanation. I think it would be good to have as a module docstring.

jesper-friis (approved):

tests/tools/conftest.py
CasperWA: This will be littered all over the code those in the future - Maybe I'll add it to the developer's documentation instead? Or do you want a comment in all module doc-strings where it'll be?

CasperWA: Added a new documentation page about testing and tools for developers :)

PR #315, opened 2021-12-01
fix #313 remove handle

### LINKED ISSUE #313: permission denied when working with temporary file
the following code creates a problem on Windows:

´´´
with tempfile.NamedTemporaryFile(suffix=".owl") as handle:
                super().save(file=handle.name, format="rdfxml", **kwargs)
                graph = rdflib.Graph()
                graph.parse(handle.name, format="xml")
                graph.serialize(destination=filename, format=format)
´´´

It is related to an issue with tempfile library. See for example discussion: https://stackoverflow.com/questions/23212435/permission-denied-to-write-to-my-temporary-file

Similar issue might be present every time ontopy tries to work with tempfile.

sygout:
# Description:

## Type of change:

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

## Checklist:

## Comments:

the following code creates a problem on Windows:

```
with tempfile.NamedTemporaryFile(suffix=".owl") as handle:
   super().save(file=handle.name, format="rdfxml", **kwargs)
   graph = rdflib.Graph() 
   graph.parse(handle.name, format="xml") 
   graph.serialize(destination=filename, format=format)
```

It is related to an issue with tempfile library. The problem comes from reopening the temporary file. the first line create and open the file. so the modification simply create a file name located in the temporary folder and use that path for the successive operations.


francescalb (approved):
Nice, works well.

CasperWA (changes requested):
The smart thing about the `with` statement and the use of `tempfile.NamedTemporaryFile()` is that the temporary file is deleted upon exiting the `with` block.
Could you possibly add this deletion manually here instead to ensure we don't clog up the temporary location?

francescalb:
> The smart thing about the `with` statement and the use of `tempfile.NamedTemporaryFile()` is that the temporary file is deleted upon exiting the `with` block. Could you possibly add this deletion manually here instead to ensure we don't clog up the temporary location?

According to the documentation we cannot use tempfile.NamedTemporaryFile() in this way on windows:
´´´
tempfile.NamedTemporaryFile(mode='w+b', buffering=-1, encoding=None, newline=None, suffix=None, prefix=None, dir=None, delete=True, *, errors=None)¶
This function operates exactly as TemporaryFile() does, except that the file is guaranteed to have a visible name in the file system (on Unix, the directory entry is not unlinked). That name can be retrieved from the name attribute of the returned file-like object. Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later). If delete is true (the default), the file is deleted as soon as it is closed. The returned object is always a file-like object whose file attribute is the underlying true file object. This file-like object can be used in a with statement, just like a normal file.
´´´

I agree with Casper about deleting the file. 

This shows us that we probably should have some windows environment tests in the CI as well...

jesper-friis:
Would it help to add the argument `mode='w'` to NamedTemporaryFile()? I think I have seen somewhere that it might help on Windows...

sygout:
According to the reference text copied by Francesca, I don't think it will help to add 'w'. 

So I will try to add the file delete.

Otherwise, yes it should be mandatory to test it on Windows, especially for a tool aimed at a larger audience.

francescalb:
> Would it help to add the argument `mode='w'` to NamedTemporaryFile()? I think I have seen somewhere that it might help on Windows...

Nope, I tested. It would maybe work with the option ´delete=False´, but then we have to delete the file afterwards as well.

CasperWA (approved):

PR #318, opened 2021-12-03
Fix checking PR body & improve error message in CD

### LINKED ISSUE #319: Edge-case fails CD workflow for dependabot
An edge-case can cause the CD workflow that updates the `ci/dependabot-updates` branch after a release to fail.
This issue was originally reported in Materials-Consortia/optimade-python-tools#995, where similar workflows are used.

CasperWA:
# Description:

There is a minor edge-case where the CD will fail updating the `ci/dependabot-updates` branch (upon a release), this minor fix sorts out that issue.
(This edge-case was found in [Materials-Consortia/optimade-python-tools](https://github.com/Materials-Consortia/optimade-python-tools) where similar workflows are used.)

Fixes #319.

It also adds a more informative error message for what steps to do next if the CD workflow fails in a particular step.

## Type of change:

- [X] Bug fix.
- [ ] New feature.
- [X] Documentation update.

## Checklist:

### DIFF
--- a/.github/workflows/ci_cd_updated_master.yml
+++ b/.github/workflows/ci_cd_updated_master.yml
@@ -33,7 +33,8 @@ jobs:
         git fetch origin
 
         LATEST_PR_BODY="$(gh api /repos/${{ github.repository}}/pulls -X GET -f state=closed -f per_page=1 -f sort=updated -f direction=desc --jq '.[].body')"
-        if [ "$(printf '%s\n' "${LATEST_PR_BODY}" | head -8)" == "$(cat .github/utils/single_dependency_pr_body.txt | head -8)" ]; then
+        cat .github/utils/single_dependency_pr_body.txt | head -8 > .tmp_file.txt
+        if [ -z "$(printf '%s\n' "${LATEST_PR_BODY}" | head -8 | diff - .tmp_file.txt --strip-trailing-cr)" ]; then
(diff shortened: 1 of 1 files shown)


francescalb:
Is there a method to test this?

CasperWA:
> Is there a method to test this?

The way I tested it originally was locally, to see which method would work - using the [gh cli](https://cli.github.com/) to download particular PR body texts and testing different ways to format it and comparisons in order to ensure that the logic here works.
Essentially the text _will_ be changed once someone clicks the checkboxes - and I think there are some other changes happening automatically if someone accidentally "edits" the body - even without changing anything. The latter is the edge case I suspect happened in the other repository, where the current solution didn't work. This one seems to have been working so far in the other repository :)

francescalb:
> > Is there a method to test this?
> 
> The way I tested it originally was locally, to see which method would work - using the [gh cli](https://cli.github.com/) to download particular PR body texts and testing different ways to format it and comparisons in order to ensure that the logic here works. Essentially the text _will_ be changed once someone clicks the checkboxes - and I think there are some other changes happening automatically if someone accidentally "edits" the body - even without changing anything. The latter is the edge case I suspect happened in the other repository, where the current solution didn't work. This one seems to have been working so far in the other repository :)

Thank you for the explanation. Then I think it is safe and good to go with this here as well-

francescalb (approved):
good

PR #327, opened 2021-12-10
Added sconverting Posix to str in get_ontology

francescalb:
# Description:

get_ontology would fail if given a valid pathlib.Posix

## Type of change:

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

## Checklist:

### DIFF
--- a/ontopy/ontology.py
+++ b/ontopy/ontology.py
@@ -84,8 +84,8 @@ class World(owlready2.World):
             of EMMO
         """
         base_iri = (
-            str(base_iri)
-            if isinstance(base_iri, pathlib.PosixPath)
+            base_iri.as_uri()
+            if isinstance(base_iri, pathlib.Path)
             else base_iri
         )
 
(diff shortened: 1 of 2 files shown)


CasperWA (commented):
Is this true only for `PosixPath` or can the more general `pathlib.Path` be used, which will "turn into" a `PosixPath` on a POSIX system and a Windows-specific `Path` on a Windows system, etc.?

Maybe one should also instead use the `as_uri()` instead of casting to a `str`?

francescalb:
> Is this true only for `PosixPath` or can the more general `pathlib.Path` be used, which will "turn into" a `PosixPath` on a POSIX system and a Windows-specific `Path` on a Windows system, etc.?
> 
> Maybe one should also instead use the `as_uri()` instead of casting to a `str`?

Good point. I ha dnot thought about windows. Indeed the whole test fails on windows already in the master ...

CasperWA:
> (...) Indeed the whole test fails on windows already in the master ...

How do you know? Did you test locally?

CasperWA (approved):
Approved on the condition that an issue is created for the Windows-issue, which may be solved by implementing a dedicated class that incorporates `pathlib.Path` as well as URLs.

PR #339, opened 2021-12-30
When visualising restrictions, annotate the edges with the restriction type by default

jesper-friis:
Used to produce the graphs for the updated README for EMMO 1.0.0-beta3:
https://github.com/emmo-repo/EMMO/tree/updated-readme-and-figures

# Description:

## Type of change:

- [ ] Bug fix.
- [x] New feature.
- [ ] Documentation update.

## Checklist:

### DIFF
--- a/ontopy/graph.py
+++ b/ontopy/graph.py
@@ -87,6 +87,7 @@ _default_style = {
         },
         "hasPart": {"color": "blue"},
         "hasProperPart": {"color": "blue", "style": "dashed"},
+        "hasMember": {"color": "blue", "style": "dotted"},
         "hasParticipant": {"color": "red"},
         "hasProperParticipant": {"color": "red", "style": "dashed"},
         "hasSpatialPart": {"color": "darkgreen"},
@@ -193,7 +194,7 @@ class OntoGraph:  # pylint: disable=too-many-instance-attributes
           - edges : attribute for individual edges (E)
         If style is None or "default", the default style is used.
         See https://www.graphviz.org/doc/info/attrs.html
-    edgelabels : bool | dict
+    edgelabels : None | bool | dict
         Whether to add labels to the edges of the generated graph.
         It is also possible to provide a dict mapping the
         full labels (with cardinality stripped off for restrictions)
@@ -231,7 +232,7 @@ class OntoGraph:  # pylint: disable=too-many-instance-attributes
         entities=None,
         relations="isA",
         style=None,
-        edgelabels=True,
+        edgelabels=None,
         addnodes=False,
         addconstructs=False,
         included_namespaces=(),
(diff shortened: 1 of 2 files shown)


jesper-friis:
The CI workflow fails with an internal [error](https://github.com/emmo-repo/EMMO-python/runs/4669272237?check_suite_focus=true#step:7:1320). Running it locally works fine:

```
$ mkdocs build --strict 
INFO     -  Cleaning site directory
INFO     -  Building documentation to directory:
            /home/friisj/prosjekter/EMMC/EMMO-python/site
INFO     -  Documentation built in 2.79 seconds
$ mkdocs --version
mkdocs, version 1.2.3 from /home/friisj/.local/lib/python3.9/site-packages/mkdocs (Python 3.9)
```

Looks more like a bug in mkdocs than in EMMOntoPy...

CasperWA (approved):
Seems fine to me - although some of it goes a bit beyond my knowledge :)

francescalb (changes requested):
image

As of now hasConstituent, hasProperPArt and has SpatialEssentialPArt are the same. As suggest that at least has Consituent and hasSpatialPart get their own colors. Furthermore, I also think it makes sense to add the label if the revert to the default color.

jesper-friis:
> As of now hasConstituent, hasProperPArt and has SpatialEssentialPArt are the same. As suggest that at least has Consituent and hasSpatialPart get their own colors. Furthermore, I also think it makes sense to add the label if the revert to the default color.

I agree that it would be a good idea to review the default colours and styles. But I think EMMO has too many relations to give all of them a unique default colour/style combination. One possibility could be to make mereological, topological and semiotical relations solid, dashed and dotted respectively. For colours we could either
- leave it to the user to define them for each plot
- define explicit colours for some high-level/important relations within each category
- define a colour cycle and more or less randomly assign them to the relations present in the graph (pros: all relations will get a unique colour by default; cons: the same relation may get different colours in two different graphs)

I think this should be a new issue.

francescalb:
> > As of now hasConstituent, hasProperPArt and has SpatialEssentialPArt are the same. As suggest that at least has Consituent and hasSpatialPart get their own colors. Furthermore, I also think it makes sense to add the label if the revert to the default color.
> 
> I agree that it would be a good idea to review the default colours and styles. But I think EMMO has too many relations to give all of them a unique default colour/style combination. One possibility could be to make mereological, topological and semiotical relations solid, dashed and dotted respectively. For colours we could either
> 
> * leave it to the user to define them for each plot
> * define explicit colours for some high-level/important relations within each category
> * define a colour cycle and more or less randomly assign them to the relations present in the graph (pros: all relations will get a unique colour by default; cons: the same relation may get different colours in two different graphs)
> 
> I think this should be a new issue.

Ok, then I think this PR can be approved.

francescalb (approved):

PR #341, opened 2022-01-02
Fix serialised ontology iri

jesper-friis:
# Description:

Serialised ontologies has IRI http://emmo.info/emmo/domain/onto.
Assigning onto.base_iri seems not to be sufficient to change that.
An easy solution is to create the ontology with the correct base IRI from the start.

## Type of change:

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

## Checklist:

### DIFF
--- a/ontopy/excelparser.py
+++ b/ontopy/excelparser.py
@@ -14,7 +14,7 @@ from typing import Tuple, Union
 import pyparsing
 import pandas as pd
 import ontopy
-from ontopy import World, get_ontology
+from ontopy import World
 from ontopy.utils import NoSuchLabelError
 from ontopy.manchester import evaluate
 import owlready2  # pylint: disable=C0411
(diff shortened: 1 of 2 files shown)


ontopy/excelparser.py
```diff
@@ -155,17 +152,14 @@ def create_ontology_from_pandas(  # pylint: disable=too-many-locals,too-many-bra
...
+    base_iri: str,
     base_iri_from_metadata: bool = True,
     catalog: dict = None,
 ) -> Tuple[ontopy.ontology.Ontology, dict]:
     """
-    Populate ontology with metada from pd.DataFrame
+    Create ontology with metadata from pd.DataFrame
     """
 
-    if onto is None:
```
francescalb, line 166: What happens now if base_iri is not given in the excelsheet? I think that the parser should not necessarily fail if it is not given

jesper-friis: Good point. I pushed a minor fix to the excelparser-update branch to also include IndexError in the except statement on line 172 to make sure that the ontology will get an default IRI rather than raising an exception.

francescalb (approved):
Seems ok

PR #344, opened 2022-01-03
Excelparser update

jesper-friis:
# Description:
Several fixes of the excel parser such that it works well with the latest version of the microstructure ontology.

## Type of change:

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

## Checklist:


ontopy/excelparser.py
```diff
@@ -9,18 +9,23 @@
...
-import sys
+from typing import Tuple, Union, Sequence
 import warnings
-from typing import Tuple, Union
-import pyparsing
+
 import pandas as pd
+import pyparsing
+
 import ontopy
```
francescalb, line 18: Is there a specific reason for the change in order of the imported packages?

ontopy/excelparser.py
```diff
@@ -9,18 +9,23 @@
...
-from ontopy import World, get_ontology
+from ontopy import get_ontology
 from ontopy.utils import NoSuchLabelError
 from ontopy.manchester import evaluate
 import owlready2  # pylint: disable=C0411
 
 
+class ExcelError(Exception):
```
francescalb, line 25:
```suggestion
class ExcelError(EMMOntoPyException):
```

ontopy/excelparser.py
```diff
@@ -9,18 +9,23 @@
...
-from typing import Tuple, Union
-import pyparsing
+
 import pandas as pd
+import pyparsing
+
 import ontopy
-from ontopy import World, get_ontology
+from ontopy import get_ontology
 from ontopy.utils import NoSuchLabelError
```
francescalb, line 20:
```suggestion
from ontopy.utils import NoSuchLabelError, EMMOntoPyException
```

francescalb (changes requested):
Can you update the onto.xlsx in examples/ontology-from-excel/tool according to the changes here? It is difficult to tedt an verify without your template.

ontopy/excelparser.py
jesper-friis: Yes there are conventions for this :-) By convention all standard libraries are imported first, followed by an empty line. Then you import other third party packages followed by yet an empty line. Finally you do the imports from the current package (which could be relative imports). Within each category I usually order them with the most fundamental packages first (according to my subjective judgement) while Casper argue that they should be ordered alphabetically.

ontopy/excelparser.py
jesper-friis: Good suggestion. Committed separately. Thanks. 

Wouldn't `EMMOntoPyError` be a better name instead of `EMMOntoPyException`?

jesper-friis:
> Can you update the onto.xlsx in examples/ontology-from-excel/tool according to the changes here? It is difficult to tedt an verify without your template.

I have been working on `tests/testonto/excelparser/onto.xlsx`. Do you prefer that I leave that file as it is and continue with `examples/ontology-from-excel/tool/onto.xlsx`? Alternatively I could move the development to https://github.com/jesper-friis/emmo-microstructure where it actually belong

ontopy/excelparser.py
francescalb: Thank you :)

ontopy/excelparser.py
francescalb: I think EMMOntoPyException makes sense to use since it inherits from Exception. Alternatively ExcelError could be renamed to ExcelException?

francescalb:
> > Can you update the onto.xlsx in examples/ontology-from-excel/tool according to the changes here? It is difficult to tedt an verify without your template.
> 
> I have been working on `tests/testonto/excelparser/onto.xlsx`. Do you prefer that I leave that file as it is and continue with `examples/ontology-from-excel/tool/onto.xlsx`? Alternatively I could move the development to https://github.com/jesper-friis/emmo-microstructure where it actually belong

Good, I think it is fine like that for now. Then we clean it up once we get to the final round before pushing to master. On master we should have one template  that is also easily accesible from the documentation, but I can fix that once we are happy with this for both the microstructure and thermodynamics example.

francescalb:
> > > Can you update the onto.xlsx in examples/ontology-from-excel/tool according to the changes here? It is difficult to tedt an verify without your template.
> > 
> > 
> > I have been working on `tests/testonto/excelparser/onto.xlsx`. Do you prefer that I leave that file as it is and continue with `examples/ontology-from-excel/tool/onto.xlsx`? Alternatively I could move the development to https://github.com/jesper-friis/emmo-microstructure where it actually belong
> 
> Good, I think it is fine like that for now. Then we clean it up once we get to the final round before pushing to master. On master we should have one template that is also easily accesible from the documentation, but I can fix that once we are happy with this for both the microstructure and thermodynamics example.

After some deliberation I think I have changed my mind. We should keep the tests/testonto/excelparser/onto.xlsx as a minimal example which includes 'Errors' and mistakes that we are testing. I rather think the file examples/ontology-from-excel/tool/onto.xlsx should be an actual real example. We can have the microstructure example there and rename it accordingly. What do you think?

jesper-friis:
> > > > Can you update the onto.xlsx in examples/ontology-from-excel/tool according to the changes here? It is difficult to tedt an verify without your template.
> > > 
> > > 
> > > I have been working on `tests/testonto/excelparser/onto.xlsx`. Do you prefer that I leave that file as it is and continue with `examples/ontology-from-excel/tool/onto.xlsx`? Alternatively I could move the development to https://github.com/jesper-friis/emmo-microstructure where it actually belong
> > 
> > 
> > Good, I think it is fine like that for now. Then we clean it up once we get to the final round before pushing to master. On master we should have one template that is also easily accesible from the documentation, but I can fix that once we are happy with this for both the microstructure and thermodynamics example.
> 
> After some deliberation I think I have changed my mind. We should keep the tests/testonto/excelparser/onto.xlsx as a minimal example which includes 'Errors' and mistakes that we are testing. I rather think the file examples/ontology-from-excel/tool/onto.xlsx should be an actual real example. We can have the microstructure example there and rename it accordingly. What do you think?

Yes, I think that make a lot of sense.

francescalb (approved):

PR #361, opened 2022-02-02
Ignore safety ID 44715 + add numpy dependency

### LINKED ISSUE #359: Add NumPy as an explicit dependency
This is to ensure a proper version control.

### LINKED ISSUE #360: Ignore NumPy `safety` warning
See this NumPy issue for more information:
https://github.com/numpy/numpy/issues/19038

Ignore safety ID 44715.

CasperWA:
# Description:

Fixes #359 
Fixes #360 

Add NumPy as an explicit dependency to be able to have version control.
Set the minimum version of NumPy to the latest supported version for Python 3.6+ (v1.19).

Ignore safety ID 44715 in CI.

## Type of change:

- [X] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

## Checklist:

- [X] Is the code easy to read and understand?
- [X] Are comments for humans to read, not computers to disregard?
- [X] Does this close the issue?
- [X] Is the change limited to the issue?
- [X] Does the new feature provide new restrictions on dependencies, and if so is this documented?

### DIFF
--- a/.github/workflows/ci_workflow.yml
+++ b/.github/workflows/ci_workflow.yml
@@ -71,8 +71,10 @@ jobs:
 
     # Ignore ID 44715 for now.
     # See this NumPy issue for more information: https://github.com/numpy/numpy/issues/19038
+    # Also ignore IDs 44716 and 44717 as they are not deemed to be as severe as it is
+    # laid out in the CVE.
     - name: Run safety
-      run: pip freeze | safety check --stdin --ignore 44715
+      run: pip freeze | safety check --stdin -i 44715 -i 44716 -i 44717
 
 
   tests:
--- a/.github/workflows/ci_workflow.yml
+++ b/.github/workflows/ci_workflow.yml
@@ -69,8 +69,10 @@ jobs:
     - name: Run PyLint
       run: pylint --rcfile=pyproject.toml *.py tools
 
+    # Ignore ID 44715 for now.
+    # See this NumPy issue for more information: https://github.com/numpy/numpy/issues/19038
     - name: Run safety
-      run: pip freeze | safety check --stdin
+      run: pip freeze | safety check --stdin --ignore 44715
 
 
   tests:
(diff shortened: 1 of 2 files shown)


CasperWA:
To be able to accommodate these `safety` issues, one will need to require minimum Python 3.8 - which is a bit drastic in my opinion. I suggest to ignore these particular `safety` issues as well, since it is suggested by the NumPy developers in the issue referenced in #360 that these CVEs are not as crucial as `safety` makes it out to be.

.github/workflows/ci_workflow.yml
```diff
@@ -69,8 +69,10 @@ jobs:
     - name: Run PyLint
       run: pylint --rcfile=pyproject.toml *.py tools
 
+    # Ignore ID 44715 for now.
+    # See this NumPy issue for more information: https://github.com/numpy/numpy/issues/19038
     - name: Run safety
-      run: pip freeze | safety check --stdin
+      run: pip freeze | safety check --stdin --ignore 44715
```
francescalb, line 75: From the last test it seems that safetly fails on both 44716 and 44717.

CasperWA: Yes, and also 44715. See my [comment above](https://github.com/emmo-repo/EMMO-python/pull/361#issuecomment-1027786041) for what I suggest for 44716 and 44717.

francescalb: Ok, I think I misunderstood you then.  I thought you had opened a finshed PR. I agree with ignoring them, if what they say about this not being a real issue is correct.

CasperWA: Yeah - it was more of an open question - because we could also go the route of dropping Python 3.6 and 3.7 support - but that seems a bit drastic.

francescalb: I agree, that is abit drastic in my eyes too.

jesper-friis (approved):
Agree with your judgement

PR #385, opened 2022-03-25
Workaround for failing test

### LINKED ISSUE #384: Pytest is currently failing
See CI jobs.

CasperWA:
# Description:

Fixes #384 

It seemed the issue here is that the used test ontology referenced URLs in it's corresponding catalog file that do not have a valid SSL certificate. Hence, the URL calls failed.
Since these URL calls are performed in Owlready2 it is difficult to avoid this or apply special configurations, other than downloading files directly in this package and pass a cached file to Owlready2.

I don't consider this PR a "fix", but rather a work-around.
Instead of truly fixing this issue and handle downloads within this package, the `"emmo"` shortcut is used in the test to use the latest (and valid) EMMO.

## Type of change:

- [X] Bug (fix).
- [ ] New feature.
- [ ] Documentation update.

## Checklist:

### DIFF
--- a/.github/workflows/cd_publish.yml
+++ b/.github/workflows/cd_publish.yml
@@ -7,7 +7,7 @@ on:
 env:
   PUBLISH_UPDATE_BRANCH: master
   GIT_USER_NAME: EMMOntoPy Developers
-  GIT_USER_EMAIL: "Team4.0@SINTEF.onmicrosoft.com"
+  GIT_USER_EMAIL: "Team4.0@SINTEF.no"
 
 jobs:
   update-and-publish:
--- a/.github/workflows/ci_automerge_dependabot.yml
+++ b/.github/workflows/ci_automerge_dependabot.yml
@@ -13,7 +13,7 @@ jobs:
     env:
       DEPENDABOT_BRANCH: ci/dependabot-updates
       GIT_USER_NAME: EMMOntoPy Developers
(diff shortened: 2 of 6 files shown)


francescalb (approved):
Very good. 
The redirection is stopped in my browser because it is http, is that related?

CasperWA:
> Very good. The redirection is stopped in my browser because it is http, is that related?

Yes - this is related. It's the same issue. It doesn't recognize / cannot validate the certificate for the site. I believe it's probably out-dated - and that presents a security risk.
So I'm reluctant to implement something that disregards this certificate check as well. These things should be respected.

On the other hand, we know the site (emmo.info) and it merely redirects to raw.githubusercontent.com and we could therefore use the latter directly. I tried that, but since the emmo.info links are given in the catalogue file, there's no way to properly use the latter only. Unless further manipulation is performed or some very specific logical choices are made that may make no sense in other use cases.
Ergo, this was my chosen workaround: Use a URL we know we can use to load an ontology to test modules.

jesper-friis:
A workaround I have used in several projects is to create my own catalog file in the folder of my application ontology with links directly to github instead of emmo.info

CasperWA:
> A workaround I have used in several projects is to create my own catalog file in the folder of my application ontology with links directly to github instead of emmo.info

That works as long as you're working locally. Here, the catalogue file is downloaded from an online location, making it immutable.

PR #389, opened 2022-03-30
[Auto-generated] Update dependencies

TEAM4-0:
### Update dependencies

Automatically created PR from [`ci/dependabot-updates`](https://github.com/emmo-repo/EMMO-python/tree/ci/dependabot-updates).

For more information see the ["Dependabot updates" workflow](https://github.com/emmo-repo/EMMO-python/blob/master/.github/workflows/ci_dependabot.yml).

#### To-do

- [x] Check that the diff is sensible, and that tests and builds pass with the new dependency versions.
- [x] Make sure that the PR is **squash** merged, with a sensible commit message.

### DIFF
--- a/.github/workflows/ci_dependabot.yml
+++ b/.github/workflows/ci_dependabot.yml
@@ -87,7 +87,7 @@ jobs:
 
     - name: Create PR
       id: cpr
-      uses: peter-evans/create-pull-request@v3
+      uses: peter-evans/create-pull-request@v4
       with:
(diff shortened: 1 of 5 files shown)


CasperWA:
As this updates Owlready2 to allow v0.37, the test failure may be due to this. If this is the case, the version should be avoided or the issue should be fixed in EMMO-python.

CasperWA (approved):

PR #396, opened 2022-04-06
create_from_excel/pandas return as list of concepts that are worngly defined in the excelfile

### LINKED ISSUE #393: Return a list of the concepts that are disregarded during when converting from excel with -force argument
Issue to respond to query from Georg Schmitz++.
This could also be a a dictionary with type of error as key and list as entry.

francescalb:
# Description:
closes #393
A dictionary with lists of concepts that are wrongly defined is returned as well as the 
produced ontology and catalog when using the --force option in create_from_excel.

The keys of the dictionary are keys on why these concepts are in the given list.

Different keys are returned for the following reasons:
* "already_defined":  These are concepts that are already in the ontology, either because they were already added in a previous line of the excelfile/pandas dataframe, or because it is already defined in the imported ontologies.
* "in_imported_ontologies": Concepts that are defined in the excel, but already exist in the imported ontologies. This is a subset of the 'already_defined'
* "wrongly_defined": Concepts that are given an invalid prefLabel (e.g. with a space in the name).
* "missing_parents": Concepts that are missing parents. These concepts are added in the top of the ontology.
* "invalid_parents": Concepts that invalidly defined parents.These concepts are added in the top of the ontology.
* "nonadded_concepts": List of all concepts that are not added, either because the prefLabel is invalid, or because the concept has already been added once or already exists in an imported ontology.

## Type of change:

- [ ] Bug fix.
- [x] New feature.
- [x] Documentation update.

## Checklist:


ontopy/excelparser.py
```diff
@@ -130,6 +158,10 @@ def create_ontology_from_pandas(  # pylint:disable=too-many-locals,too-many-bran
     onto, catalog = get_metadata_from_dataframe(
         metadata, base_iri, imports=imports
     )
+    # Get a set of imported concepts
+    imported_concepts = {
+        concept.prefLabel.first() for concept in onto.get_by_label_all("*")
```
jesper-friis, line 163: get_entities() might be slightly more efficient to use.

PR #399, opened 2022-04-10
Add functionality for setting name part of IRI to prefLabel

### LINKED ISSUE #398: Add functionality for setting name part of IRI to prefLabel

jesper-friis:
fix #398 

# Description:
Added `--rename-iris` and `--annotate-source` options of ontoconvert.
In order to allow combining these options with with the `--squash`, `--recursive` and `--reasoner` options, the `--squash` and `--recursive` functionality had to be moved into Ontology.save(), which made this PR ending up being quite large... It didn't became better, when I realised that the Owlready2 currently completely messes up the ontology when adding new relations from the reasoner...

Changes:
* implemented ontopy.utils.rename_iris()
* implemented ontopy.utils.annotate_sources()
* improved writing of catalog files
* added `recursive`, `squash` and `write_catalog_file` keyword arguments to Ontology.save()
* cleaned up Ontology.sync_reasoner()
* added new options to the ontoconvert tool
* added a workaround to ontoconvert for correct reasoning using rdflib - if the related bug in Owlready2 is not fixed soon, we should consider to add a correct implementation of owlready2.reasoning._apply_reasoning_results() to EMMOntoPy.

## Type of change:

- [ ] Bug fix.
- [ x] New feature.
- [ ] Documentation update.

## Checklist:


tools/ontoconvert
```diff
@@ -111,53 +182,129 @@ def main(argv: list = None):
...
+
+        if args.rename_iris:
+            rename_iris(onto, args.rename_iris)
+
+        # owlready2.reasoning._apply_reasoning_results() is broken!
+        # Currently it messes up everything by adding the inferred
+        # rdfs:subClassOf relations as rdf:type relations :-(
+        #
+        # Temporary workaround
+        # Instead of calling onto.sync_reasoner() we call FaCTPPGraph
```
francescalb, line 202: This should be inccluded in the documentation

tools/ontoconvert
```diff
@@ -111,53 +182,129 @@ def main(argv: list = None):
...
+        #     format=output_format,
+        #     dir=args.output_dir,
+        #     mkdir=True,
+        #     overwrite=args.overwrite,
+        #     recursive=args.recursive,
+        #     squash=args.squash,
+        #     write_catalog_file=bool(args.catalog_file),
+        #     append_catalog=args.append_catalog,
+        #     catalog_file=args.catalog_file,
+        # )
```
francescalb, line 307: Why having all this commented out?

francescalb (changes requested):
Remove uneccesary functions that remain in utils so that we do not do the same things wtice in two different places.

francescalb (changes requested):
Also a test is needed

tools/ontoconvert
jesper-friis: Because we want to bring back these lines again when the bug in sync_reasoner() has been fixed.

francescalb (approved):
Looks good. I think the has_triple and get_triples can be made as general utility functions. I suggest adding an Issue on that referring to the code here.

PR #405, opened 2022-04-26
Flb/get descendants

francescalb:
# Description:
Added get_descendants as function to ontology
* takes arguments
	- classes: one owl.ThingClass or an iterable sequence of such.
	- common: wether to return only common descendants or not
	- generations: number of generations to return. Defaults to all.

## Type of change:

- [ ] Bug fix.
- [x] New feature.
- [ ] Documentation update.

## Checklist:

### DIFF
--- a/ontopy/ontology.py
+++ b/ontopy/ontology.py
@@ -9,6 +9,7 @@ The class extension is defined within.
 If desirable some of this may be moved back into owlready2.
 """
 # pylint: disable=too-many-lines,fixme,arguments-differ,protected-access
+from typing import TYPE_CHECKING, Union
 import os
 import itertools
 import inspect
@@ -17,7 +18,6 @@ import uuid
 import tempfile
 import types
 import pathlib
-from typing import Union
 from collections import defaultdict
 from collections.abc import Iterable
 
@@ -46,6 +46,9 @@ from ontopy.utils import (
 )
 
 from ontopy.ontograph import OntoGraph  # FIXME: deprecate...
+
+if TYPE_CHECKING:
+    from typing import List
 
 
 # Default annotations to look up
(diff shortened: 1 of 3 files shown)


ontopy/ontology.py
```diff
@@ -1270,6 +1273,60 @@ def addancestors(entity, counter, subject):
...
+        - all_descendants: Include all descendants.
+        Returns:
+            A list of descendants including required generation.
+            If 'common'=True, the common descendants are returned
+            within the specified number of generations.
+            If 'generations' is not given and 'common' is True all
+            descendants will be checked.
+            'generations' defaults to 1 if 'common' not True.
+        """
+        if (common) & (generations == -1):
```
jesper-friis, line 1297:
```suggestion
        if common and generations == -1:
```

ontopy/ontology.py
```diff
@@ -1270,6 +1273,60 @@ def addancestors(entity, counter, subject):
...
+        self,
+        classes: "Union[List, ThingClass]",
+        common: bool = False,
+        generations: int = -1,
+        all_descendants: bool = False,
+    ):
+        """Return descendants/subclasses of all classes in `classes`.
+        Arguments:
+        - classes: to be provided as list.
+        - common: only return common descendants.
```
jesper-friis, line 1286:
```suggestion
        - common: whether to only return descendants common to all classes.
```

ontopy/ontology.py
```diff
@@ -1270,6 +1273,60 @@ def addancestors(entity, counter, subject):
             return ancestors.difference(classes)
         return ancestors
 
+    def get_descendants(
+        self,
+        classes: "Union[List, ThingClass]",
+        common: bool = False,
+        generations: int = -1,
+        all_descendants: bool = False,
```
jesper-friis, line 1281: Personally I find a single option with a fill value simpler than two closely coupled options. 

I would therefore have removed `all_descendants` and made `generations` equal -1 or (better) None mean to include all generations in the output. Since there are no obvious reason for a particular default value for `generations`, I would have made returning all descendants the default.

However, all of this is of cause a matter of taste, which you can take or leave as you wish.

ontopy/ontology.py
```diff
@@ -1270,6 +1273,60 @@ def addancestors(entity, counter, subject):
...
+            If 'common'=True, the common descendants are returned
+            within the specified number of generations.
+            If 'generations' is not given and 'common' is True all
+            descendants will be checked.
+            'generations' defaults to 1 if 'common' not True.
+        """
+        if (common) & (generations == -1):
+            all_descendants = True
+        elif generations == -1:
+            generations = 1
```
jesper-friis, line 1300: Is there a particular reason for this logic? I would suggest to remove it, to decouple the arguments from each other.

ontopy/ontology.py
```diff
@@ -1270,6 +1273,60 @@ def addancestors(entity, counter, subject):
...
+                if num < generations:
+                    _children_recursively(num + 1, child, parent, descendants)
+
+        if generations == 0:
+            return set()
+
+        if all_descendants is True:
+            for entity in classes:
+                descendants[entity] = entity.descendants()
+
```
jesper-friis, line 1320:
```suggestion
```

ontopy/ontology.py
```diff
@@ -1270,6 +1273,60 @@ def addancestors(entity, counter, subject):
...
+        else:
+            for entity in classes:
+                _children_recursively(1, entity, entity, descendants)
+
+        results = [val for _, val in descendants.items()]
```
jesper-friis, line 1325:
```suggestion
        results = [val for val in descendants.values()]
```

ontopy/ontology.py
```diff
@@ -1270,6 +1273,60 @@ def addancestors(entity, counter, subject):
...
+        all_descendants: bool = False,
+    ):
+        """Return descendants/subclasses of all classes in `classes`.
+        Arguments:
+        - classes: to be provided as list.
+        - common: only return common descendants.
+        - generations: Include this number of descendant levels.
+        - all_descendants: Include all descendants.
+        Returns:
+            A list of descendants including required generation.
```
jesper-friis, line 1290:
```suggestion
            A set of descendants including required generation.
```

ontopy/ontology.py
```diff
@@ -1270,6 +1273,60 @@ def addancestors(entity, counter, subject):
             return ancestors.difference(classes)
         return ancestors
 
+    def get_descendants(
+        self,
+        classes: "Union[List, ThingClass]",
+        common: bool = False,
+        generations: int = -1,
+        all_descendants: bool = False,
+    ):
```
jesper-friis, line 1282:
```suggestion
    ) -> set:
```

ontopy/ontology.py
```diff
@@ -1270,6 +1273,60 @@ def addancestors(entity, counter, subject):
...
+            If 'generations' is not given and 'common' is True all
+            descendants will be checked.
+            'generations' defaults to 1 if 'common' not True.
+        """
+        if (common) & (generations == -1):
+            all_descendants = True
+        elif generations == -1:
+            generations = 1
+
+        if not isinstance(classes, list):
```
jesper-friis, line 1302: Better to check for any type of sequence with
```
        if not isinstance(classes, typing.Sequence):
```

ontopy/ontology.py
francescalb: I was thinking to have it different from entity.descendants as we  initally wanted this to get the first generation. But I agree with you that it is better to have all as default.

ontopy/ontology.py
francescalb: Because if we ask for only common descendants it is more likely that we are interested in all generations. But if we change default to all generations this goes out anyhow. I'll remove it.

ontopy/ontology.py
jesper-friis: It is possible to simplify the logics by skipping the temporary dict by something like the following: 
```python
def get_descendants(self, classes, common=False, generations=None):
    if not isinstance(classes, Sequence):
        classes = [classes]

    def get_children(parent, children, n=0):
        if generations is not None and n < generations:
            for child in self.get_children_of(parent):
                children.add(child)
                get_children(child, children, n+1)
        return children

    children = get_children(classes[0], set())
    for cls in classes[1:]:
        if common:
            children.intersection_update(get_children(cls, set()))
        else:
            children.update(get_children(cls, set()))

    return children
```

jesper-friis (commented):
Added some comments. Good test

jesper-friis (approved):

PR #446, opened 2022-08-19
Fix loading in windows, url paths

francescalb:
This works for me now in both windows and linux.
onto = get_ontology('https://raw.githubusercontent.com/BIG-MAP/BattINFO/master/battinfo.ttl').load()**

Can you please check?

# Description:

## Type of change:

- [ ] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

## Checklist:

### DIFF
--- a/ontopy/utils.py
+++ b/ontopy/utils.py
@@ -350,7 +350,7 @@ def read_catalog(  # pylint: disable=too-many-locals,too-many-statements,too-man
                 url = f"{baseuri}/{uri_as_str}"
             else:
                 url = os.path.join(baseuri if baseuri else dirname, uri_as_str)
-        url = normalise_url(url)
+        # url = normalise_url(url)
 
         iris.setdefault(uri.attrib["name"], url)
         if recursive:


ontopy/utils.py
```diff
@@ -350,7 +350,7 @@ def load_uri(uri, dirname):
                 url = f"{baseuri}/{uri_as_str}"
             else:
                 url = os.path.join(baseuri if baseuri else dirname, uri_as_str)
-        url = normalise_url(url)
+        # url = normalise_url(url)
```
CasperWA, line 353: Wouldn't this create other issues, since the URL is now not normalized?

francescalb: Both Jesper and I tried to see if we found such cases, but did not. The reason why it was commented out and not deleted completely was for easy reversal (and then searching for a new solution for Windows) if needed.

sygout:
I tried to run pytest on Windows and got: 5 failed, 17 passed, 2 skipped, 7 warnings

francescalb:
> I tried to run pytest on Windows and got: 5 failed, 17 passed, 2 skipped, 7 warnings

Yes, I do to on windows. We need to set up a testing environment for windows. Since I (almost) never use windows for this I have not encoutnered them. I suggest making a separate issue for creating a windows testing environment though. And we just make sure that loading works in both windows and linux for now

sygout:
On a fresh installation (new virtual environment and clean repository clone), the correction seems to work as test_load is working, the error present in master v0.3.1-19-g02b3c84 (FAILED test_load.py::test_load - urllib.error.HTTPError: HTTP Error 400: https://raw.githubusercontent.com/%5CBIG-MAP) does not appear anymore.

PR #470, opened 2022-09-27
Setting prefix explicitly in excelparser

francescalb:
# Description:
Prefix can now be set explicitly for imported ontologies in excelparser. Either only prefix of given ontology, or set common prefix for all ontologies with a given root in the base_iri. Importing subontologies is also corrected.

The following is an image of the ImportedOntologies excelsheet. 
image

The following image shows two new tests added in the excelparser test, in which two new classes are added. Both of these are subclasses on the locally imported testontology, either from the first import or its subontology. Note that they are given the same prefix based on the base_iri_root. If base_iri_root is not given the given prefix is only set to the main ontology and not the indirect imports. If base_iri_root is given all ontologies with their base_iri starting with base_iri_root are given the same prefix.

image

## Type of change:

- [ ] Bug fix.
- [x] New feature.
- [ ] Documentation update.

## Checklist:


ontopy/excelparser.py
```diff
@@ -99,18 +100,19 @@ def create_ontology_from_excel(  # pylint: disable=too-many-arguments
...
         )
     except ValueError:
         pass
     else:
-        # Strip leading and trailing white spaces in path
-        imports.extend(
-            imports_frame["Imported ontologies"].str.strip().to_list()
-        )
+        # Strip leading and trailing white spaces in paths
+        imports = imports.replace(" ", "")
```
jesper-friis, line 113: This removes all spaces in path. Use `imports.strip()` to strip leading and trailing white spaces.

francescalb: Good point. I did not consider the possibility of adding the path to a local file with spaces in the path.

jesper-friis (approved):
Looks very nice

CasperWA:
@francescalb I think this can be merged, and then you can try to perform a release again using all the new bells and whistles :)

PR #472, opened 2022-09-28
Ensure that generated IRIs are unique but cosistent.

jesper-friis:
Base generated IRIs on both `base_iri` and `name` to ensure that they will differ in two different ontologies that happens to define the same name.

## Type of change:
- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

## Checklist:


jesper-friis:
Since the `base_iri` will be prepended to the generated UUID, this change should not make a difference, except of changing all generated IRIs.

Maybe this PR should be rejected?

PR #507, opened 2022-11-25
Catch TypeError if `is_instance_of` is not iterable

### LINKED ISSUE #506: ontodoc: Expect `is_instance_of` property to be iterable
Trying to run `ontodoc` for the new [domain-battery](https://github.com/emmo-repo/domain-battery) locally ends up with a bug in `ontodoc.py` where it is running through entities, checking instances in it and iterating over an item's `is_instance_of` property, expecting it to be iterable. However, in the local case, it is not, which raises a `TypeError`.
This can be remedied by introducing a `try/except` test, passing for `TypeError`.

CasperWA:
# Description:

Fixes #506 

See the issue for more information.

Unfortunately, I am not aware of how the `is_instance_of` property is generated or how to create a test ontology to create a test for this particular example that would trigger the bug prior to the current change.

## Type of change:

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

## Checklist:

### DIFF
--- a/ontopy/ontodoc.py
+++ b/ontopy/ontodoc.py
@@ -391,9 +391,15 @@ class OntoDoc:
         # Instances (individuals)
         if hasattr(item, "instances"):
             points = []
-            for entity in [
-                _ for _ in item.instances() if item in _.is_instance_of
-            ]:
+            entities = []
+            for entity in item.instances():
+                try:
+                    if item in entity.is_instance_of:
+                        entities.append(entity)
+                except TypeError:
+                    # entity.is_instance_of is not an iterable
+                    pass
+            for entity in entities:
                 points.append(
                     point_style.format(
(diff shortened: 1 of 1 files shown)


francescalb (commented):
Can you provide the example that failed? I am trying to run ontodoc locally on the battinfo ontology, but get a different error.

francescalb:
> Can you provide the example that failed? I am trying to run ontodoc locally on the battinfo ontology, but get a different error.

The error was generated for me now. I'll look into it.

francescalb:
After some digging, I foudn out that this error is thrown because of punning. After discussions with Simon (main author of the battery-ontology), Emanuele (main author of EMMO) and Jesper, we agreed that it is OK to not support punning for ontodoc at the present moment. I will open a new PR for this, as per agreed at stand-up.

PR #509, opened 2022-12-08
Do not trigger an emmocheck failure for ontologies with a foaf:logo annotation

jesper-friis:
CHAMIO (and EMMO with PR [212](https://github.com/emmo-repo/EMMO/pull/212)) includes a foaf:logo annotation.
This PR makes sure that this does not trigger a missing prefLabel error.

**Note** that the "automatic" prefix assigned by OwlReady2 for foaf is rather awkward. It becomes "0.1" since that is the last component of the foaf IRI: `http://xmlns.com/foaf/0.1/`. 

# Description:

## Type of change:

- [ ] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

## Checklist:

### DIFF
--- a/emmopy/emmocheck.py
+++ b/emmopy/emmocheck.py
@@ -82,6 +82,7 @@ class TestSyntacticEMMOConventions(TestEMMOConventions):
                 "core.prefLabel",
                 "core.altLabel",
                 "core.hiddenLabel",
+                "0.1.logo",  # foaf.logo
             )
         )
         exceptions.update(


requirements_dev.txt
```diff
@@ -2,3 +2,4 @@ pre-commit~=2.20
 pylint~=2.15
 pytest~=7.2
 pytest-cov~=4.0
+setuptools~=57.4
```
francescalb, line 5: Why? We have not needed setuptools until now, why add this in the current PR, and why then in _dev?

francescalb (commented):
PRs should be made as a response to an issue, which is then closed with the PR.

requirements_dev.txt
jesper-friis: I have tested around changing different requirements without any success. This didn't worked, so it can be removed again...

francescalb (approved):
For some reason the action has stoppd and is not in the queue , but this should be fine to merge now.

Issue #511, opened 2022-12-12
If there are altLabels that match, get_by_label_all returns only the prefLabels.


francescalb:
Example:
´´´
from ontopy import get_ontology
onto = get_ontology()
onto = get_ontology().load()
onto.get_by_label_all('*Concentration*')
´´´

How to solve this should be consdered. Alternative labels should be returned as well, I think, but I am not sure if we should specifiy which are altLabels and which are prefLabels. 

When this is fixed test_prefix should be corrected.

PR #512, opened 2022-12-12
Added fix for adding annotations that are not strings in ontodoc

francescalb:
Furthermore, reenabled test for ontodoc. This is still a minimal test, but I think it can be gradually expanded.

Lastly - because a new altLabel was added in the testontology, it was reveiled that the same entity is returned twice in get_by_label_all in test_prefix. Issue #511 was made bacause of this.

# Description:

## Type of change:

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

## Checklist:

### DIFF
--- a/ontopy/ontodoc.py
+++ b/ontopy/ontodoc.py
@@ -288,7 +288,8 @@ class OntoDoc:
             annotations.keys(), key=lambda key: order.get(key, key)
         ):
             for value in annotations[key]:
-                if self.url_regex.match(value):
+                value = str(value)
+                if self.url_regex.match(str(value)):
                     doc.append(
                         annotation_style.format(
                             key=key, value=asstring(value, link_style)
--- a/ontopy/ontodoc.py
+++ b/ontopy/ontodoc.py
@@ -289,7 +289,7 @@ class OntoDoc:
         ):
             for value in annotations[key]:
                 value = str(value)
-                if self.url_regex.match(str(value)):
+                if self.url_regex.match(value):
                     doc.append(
                         annotation_style.format(
                             key=key, value=asstring(value, link_style)
--- a/tests/testonto/models.ttl
+++ b/tests/testonto/models.ttl
@@ -18,4 +18,5 @@ skos:altLabel rdf:type owl:AnnotationProperty .
 
 :testclass rdf:type owl:Class ;
     rdfs:subClassOf owl:Thing ;
-    skos:prefLabel "TestClass"@en .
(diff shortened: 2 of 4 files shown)


tests/testonto/models.ttl
```diff
@@ -18,4 +18,5 @@ skos:altLabel rdf:type owl:AnnotationProperty .
 
 :testclass rdf:type owl:Class ;
     rdfs:subClassOf owl:Thing ;
-    skos:prefLabel "TestClass"@en .
+    skos:prefLabel "TestClass"@en ;
+    skos:altLabel 25517 . # Test that values given as integers as still accepted by ontodoc
```
francescalb, line 22:
```suggestion
    skos:altLabel 25517 . # Test that values given as integers are accepted by ontodoc
```

ontopy/ontodoc.py
```diff
@@ -288,7 +288,8 @@ def itemdoc(
             annotations.keys(), key=lambda key: order.get(key, key)
         ):
             for value in annotations[key]:
-                if self.url_regex.match(value):
+                value = str(value)
+                if self.url_regex.match(str(value)):
```
francescalb, line 292:
```suggestion
                if self.url_regex.match(value):
```

ontopy/ontodoc.py
```diff
@@ -288,6 +288,7 @@ def itemdoc(
             annotations.keys(), key=lambda key: order.get(key, key)
         ):
             for value in annotations[key]:
+                value = str(value)
```
CasperWA, line 291: This could fail, I suppose, if the type of `value` cannot be cast to a `str` type.
You could instead wrap it in a `try/except` block, where you're catching `TypeError` and `ValueError` exceptions, re-raising them with an appropriate custom message?

CasperWA (changes requested):

ontopy/ontodoc.py
francescalb: I discussed this with @jesper-friis , and we can't really see a situation in which this not enough. I think it is better to leave it as is, and update this if and when we find examples of situations in which a different handling is needed.

CasperWA: As you wish. Casting to `str` is in general quite generous in this case, so it's fine I suppose. However, I'd argue it's good convention to consider possible edge cases and try to catch them if possible so exception messages are clear and helpful.

CasperWA (approved):

PR #515, opened 2022-12-12
Removed LegacyVersion from ontopy.utils

francescalb:
# Description:

## Type of change:

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

## Checklist:

### DIFF
--- a/ontopy/utils.py
+++ b/ontopy/utils.py
@@ -450,7 +450,7 @@ def _validate_installed_version(
 
     if isinstance(min_version, str):
         min_version = parse_version(min_version)
-    elif isinstance(min_version, (Version)):
+    elif isinstance(min_version, Version):
         # We have the format we want
         pass
     else:
--- a/ontopy/utils.py
+++ b/ontopy/utils.py
@@ -12,6 +12,7 @@ import urllib.request
 import urllib.parse
 import warnings
 import defusedxml.ElementTree as ET
+from packaging.version import parse as parse_version, Version
 
 from rdflib import Graph, URIRef
 from rdflib.util import guess_format
@@ -21,7 +22,6 @@ import owlready2
 
 
 if TYPE_CHECKING:
-    from packaging.version import Version
     from typing import Optional, Union
 
 
@@ -447,7 +447,6 @@ def _validate_installed_version(
     """
     # pylint: disable=import-outside-toplevel
     import importlib
-    from packaging.version import parse as parse_version, Version
 
     if isinstance(min_version, str):
         min_version = parse_version(min_version)
--- a/ontopy/utils.py
+++ b/ontopy/utils.py
@@ -21,7 +21,7 @@ import owlready2
 
 
 if TYPE_CHECKING:
-    from packaging.version import Version, LegacyVersion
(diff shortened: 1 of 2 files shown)


francescalb:
LegacyVersion was removed from the package 'packaging'. It was noly used for type_checking, so removing should not be a problem.

CasperWA (commented):
I don't think there is a need to remove `LegacyVersion` here?
Or does it not exist any more in the newest version of `packaging`?

CasperWA:
> I don't think there is a need to remove `LegacyVersion` here? Or does it not exist any more in the newest version of `packaging`?

Ah. I should just read your comment. Didn't find a reason for the PR at first because it wasn't in the PR description but buried underneath, sorry.

ontopy/utils.py
```diff
@@ -447,16 +447,16 @@ def _validate_installed_version(
...
     if isinstance(min_version, str):
         min_version = parse_version(min_version)
-    elif isinstance(min_version, (LegacyVersion, Version)):
+    elif isinstance(min_version, (Version)):
         # We have the format we want
         pass
     else:
         raise TypeError(
-            "min_version should be either a str, LegacyVersion or Version. "
+            "min_version should be either a str (Version). "
```
CasperWA, line 459:
```suggestion
            "min_version should be either a str or packaging.Version. "
```

ontopy/utils.py
```diff
@@ -447,16 +447,16 @@ def _validate_installed_version(
     """
     # pylint: disable=import-outside-toplevel
     import importlib
-    from packaging.version import parse as parse_version, LegacyVersion, Version
+    from packaging.version import parse as parse_version, Version
 
     if isinstance(min_version, str):
         min_version = parse_version(min_version)
-    elif isinstance(min_version, (LegacyVersion, Version)):
+    elif isinstance(min_version, (Version)):
```
CasperWA, line 454:
```suggestion
    elif isinstance(min_version, Version):
```
This is not going to work if `Version` is only imported in the `TYPE_CHECKING` block!

ontopy/utils.py
```diff
@@ -21,7 +21,7 @@
 
 
 if TYPE_CHECKING:
-    from packaging.version import Version, LegacyVersion
+    from packaging.version import Version
```
CasperWA, line 24: Move this up and out of the `TYPE_CHECKING` block as `Version` is used explicitly in line 454 and not just as a type hint.

ontopy/utils.py
```diff
@@ -447,16 +447,16 @@ def _validate_installed_version(
     """
     # pylint: disable=import-outside-toplevel
     import importlib
-    from packaging.version import parse as parse_version, LegacyVersion, Version
+    from packaging.version import parse as parse_version, Version
```
CasperWA, line 450:
```suggestion
    from packaging.version import parse as parse_version
```

CasperWA: Import `Version` and the top of the file instead.

CasperWA (changes requested):

CasperWA (approved):

PR #526, opened 2023-01-19
Only generate new uuid if not already a valid one

francescalb:
# Description:
When asking to sync_attributes with the uuid name_policy the name is regenerated for all, even if the alrealy have a valid name within this name_policy. This makes it impossible to add to an already existing ontology.

## Type of change:

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

## Checklist:


jesper-friis (approved):
I will suggest to check that `obj.name` starts with `name_prefix`. Except for that I think this is the right fix. 

I approve to not cause unnecessarily delays.

PR #527, opened 2023-01-19
Added possibility to update ontology. 

francescalb:
# Description:
If a file with name as the output ontology already exists it will automatically load this ontology and only add new concepts that are not already defined. 

Updating the ontology instead of regenerating is the default, but can be overridden with -update_ontology= False. A completely new ontology will then be regenerated. 

for excel2onto: If input_ontologyis not given it will take this as equal to outputname, but can also be specified to be different.

## Type of change:

- [ ] Bug fix.
- [x] New feature.
- [ ] Documentation update.

## Checklist:

- [x] Is the code easy to read and understand?
- [x] Are comments for humans to read, not computers to disregard?
- [x] Does a new feature has an accompanying new test (in the CI or unit testing schemes)?
- [x] Has the documentation been updated as necessary?
- [x] Does this close the issue?
- [x] Is the change limited to the issue?
- [x] Are errors handled for all outcomes?
- [x] Does the new feature provide new restrictions on dependencies, and if so is this documented?


lovfall (approved):
Looks good to me

PR #532, opened 2023-02-02
Bypass punning in ontodoc.

### LINKED ISSUE #506: ontodoc: Expect `is_instance_of` property to be iterable

francescalb:
# Description:
Closes #506.
Ontodoc fails when punned individuals are included in the ontology. 
After discussions with various people (see #507)  we have agreed to not support punning for now.
This might be changed in the future if we see the need.

Tests have been added to see that ontodoc runs for ontologies with an individual, and that it runs with a warning
when the individual is a punned individual. Pytest is set to ignore that particualr warning so that this test 
will continue to pass if pytest is set to fail on warnings as well in the future.
 
## Type of change:

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

## Checklist:

### DIFF
--- a/ontopy/ontodoc.py
+++ b/ontopy/ontodoc.py
@@ -393,28 +393,28 @@ class OntoDoc:
         if hasattr(item, "instances"):
             points = []
             for instance in item.instances():
-                try:
-                    if item in instance.is_instance_of:
-                        points.append(
-                            point_style.format(
-                                point=asstring(instance, link_style),
-                                ontology=onto,
-                            )
-                        )
-                    if points:
-                        value = points_style.format(
-                            points="".join(points), ontology=onto
-                        )
-                        doc.append(
-                            annotation_style.format(
-                                key="Individuals", value=value, ontology=onto
-                            )
-                        )
-                except TypeError:
+                if isinstance(instance.is_instance_of, property):
                     warnings.warn(
                         f'Ignoring instance "{instance}" which is both and '
                         "indivudual and class. Ontodoc does not support "
                         "punning at the present moment."
+                    )
+                    continue
+                if item in instance.is_instance_of:
+                    points.append(
+                        point_style.format(
+                            point=asstring(instance, link_style),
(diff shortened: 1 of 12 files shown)


ontopy/ontodoc.py
```diff
@@ -392,23 +392,30 @@ def itemdoc(
...
+                        value = points_style.format(
+                            points="".join(points), ontology=onto
+                        )
+                        doc.append(
+                            annotation_style.format(
+                                key="Individuals", value=value, ontology=onto
+                            )
+                        )
+                except TypeError:
+                    warnings.warn(
```
jesper-friis, line 414: Very good warning message. Are we sure that there are not other TypeErrors what will be catched by this except statement?

jesper-friis (approved):
Looks good. Added a question

ontopy/ontodoc.py
francescalb: No, I am not sure that there are not others. But I think it is only punning that will be caught by this because we ask for an indivudal, and then we iterate of the concepts that the individual is an instance of. But I think this is a good point. If there is punning I know that is_instance_of returns 'property' and not a list of concepts, so I changed it to be a positive test for that. Better?

jesper-friis (approved):
Yes, this looks very good. 

I have not tested it locally which I normally would do if I had my linux machine at hand, but I see you have a good test.

As far as I can see, PR #535 fixes the issue with failing tests. Please merge that into this branch before merging to master.

jesper-friis (approved):
Accepted given that the test passes

PR #536, opened 2023-02-03
Fix fixtures for Python3.7

CasperWA:
# Description:

Fix issue with fancy pytest fixtures for running the tools as a Python module for Python 3.7.

Basically, the fixtures have been reimplemented by splitting the single `tool` fixture into two separate fixtures:
- `get_tool`: Import a tool as a module.
  Essentially, one uses this fixture in a test and then runs `ontodoc = get_tool("ontodoc")` to import the `tools/ontodoc` file as a module. It can now be tested by running its `main()` function.
- `rename_tools`: Add a `.py` extension to all tools.
  This fixture runs prior to **all** tests in this module (the `tests/tools/` folder). First, it adds a `.py` suffix to all the tool file names to make them importable as a module. Then it stops executing for a while (yields) and then all the tests can be run.
  Then after they're done it reverts the file names, by removing the `.py` suffix. Furthermore, it returns `sys.path` to its original state prior to running the tests; to make the imports of the tools work, the `tools` folder had to be added to the `sys.path`.

I updated the doc-strings for all the tests to be more clear about where the fixtures are coming from as well, and tried to minimize the imports outside the test functions as much as possible.

## Type of change:

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

## Checklist:

### DIFF
--- a/.github/workflows/ci_workflow.yml
+++ b/.github/workflows/ci_workflow.yml
@@ -1,63 +1,63 @@
 name: CI Tests
 
 on:
-  pull_request:
+  # pull_request:
   push:
-    branches:
-      - master
-      - 'push-action/**'
+    # branches:
+    #   - master
+    #   - 'push-action/**'
 
 jobs:
-  tests:
-    name: External
-    uses: SINTEF/ci-cd/.github/workflows/ci_tests.yml@v1
-    with:
-      # General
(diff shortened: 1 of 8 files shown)


francescalb (approved):
Excellent!

PR #548, opened 2023-02-07
Fix internal links in generated documentation generated with ontodoc

jesper-friis:
# Description:
PR #543 changed the id of documented concepts from preflabel to the name-part of the IRI. This PR updates internal links accordingly such that they works again.

The implementation became little more involved, because `utils.asstring()` now needs access to the ontology.

## Type of change:
- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

## Checklist:


francescalb (changes requested):
This new version does not write out the iris correctly:
image

As you see, it give IRI: Entropy. The link is the IRI but written as prefLabel. This seems wrong to me. Also, didn't it used to be that one can click from one concept in the document to another? Now all clicking leads to the same https://raw.githubusercontent.com/emmo-repo/EMMO/master/emmo.ttl#IRI ( IRI being anythin as it is not in that file).

jesper-friis:
> This new version does not write out the iris correctly: image
> 
> As you see, it give IRI: Entropy. The link is the IRI but written as prefLabel. This seems wrong to me. Also, didn't it used to be that one can click from one concept in the document to another? Now all clicking leads to the same https://raw.githubusercontent.com/emmo-repo/EMMO/master/emmo.ttl#IRI ( IRI being anythin as it is not in that file).

Very well seen! Thanks. That the IRIs ended up being shown with their prefLabel was a logical error, which is solved now.
The `asstring()` function now treats a string argument as the string to show. It still tries to be smart and see if it can infer an entity from the string argument, but that will only affect the link, not the shown value.

The other issue with the links was actually by design (as requested by BIG-MAP). However, the old behaviour with relative links make much more sense, especially before the redirections works properly. Hence, we are now generating relative links by default.

francescalb (approved):
Looks good.

PR #561, opened 2023-03-06
Corrected bug on getting default relation style.

francescalb:
# Description:
A new relation without a default style was used in the test of ontograph, which showed that there was an error in this part.
A new test has been added that will eventually cover this, by accepting only that one warning, i.e. when runnig pytest -Werror this test will pass. 

## Type of change:

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

## Checklist:


ontopy/graph.py
```diff
@@ -714,7 +714,7 @@ def get_edge_attrs(self, predicate, attrs):
                     self.ontology[_] for _ in relations if _ in self.ontology
                 )
                 for relation in entity.mro():
-                    if relation in rels:
+                    if relation in relations:
                         rattrs = (
                             relations[get_label(relation)]
                             if relation in rels
```
CasperWA, line 720: Sure this is `relations` and not `rels`, what's the difference between these? (Defined in lines 712 and 713.)

ontopy/utils.py
```diff
@@ -109,7 +109,7 @@ def getiriname(iri):
 
 def asstring(  # pylint: disable=too-many-return-statements,too-many-branches,too-many-statements
     expr,
-    link="{lowerlabel}",
+    link="{label}",
```
CasperWA, line 112: Shouldn't the link still be "the label of the concept in lower case and with spaces replaced with hyphens"?

tests/ontopy_tests/test_graph2.py
```diff
@@ -1,14 +1,43 @@
 from typing import TYPE_CHECKING
+import owlready2
+import pytest
+from ontopy.graph import OntoGraph
```
CasperWA, line 4: Move these imports into the function for better test isolation during runtime.

tests/ontopy_tests/test_graph2.py
```diff
@@ -1,14 +1,43 @@
...
     from pathlib import Path
 
     from ontopy.ontology import Ontology
 
 
-def test_graph2(emmo: "Ontology", tmpdir: "Path") -> None:
-    from ontopy.graph import OntoGraph
+@pytest.mark.filterwarnings(
+    "ignore:Style not defined for relation hasSpecialRelation."
+)  # currently pytest is set to accept warnings, but this might change in the future
```
CasperWA, line 14: Is it important for the test the warning is ignored?
Otherwise, I don't think it's relevant to ignore here. Rather one can either:

- Move the ignore statement to the general pytest configuration (`pyproject.toml`):

  ```ini
  [tool.pytest.ini_options]
  filterwarnings = [
      # Some comment here on why this is ignored.
      "ignore:Style not defined for relation hasSpecialRelation.",
  ]
  ```

- Remove the filterwarning ignore statement altogether.

tests/ontopy_tests/test_graph2.py
```diff
@@ -1,14 +1,43 @@
...
+def test_graph(testonto: "Ontology", tmpdir: "Path") -> None:
+    with testonto:
```
CasperWA, line 16:
```suggestion
def test_graph(testonto: "Ontology", tmpdir: "Path") -> None:
    """Short description of the test.

    Parameters:
        testonto: A local fixture from the root `conftest.py`.
            Returns a local ontology used for testing.
        tmpdir: A built-in pytest fixture to get a function-scoped
            temporary directory.

    """
    with testonto:
```

@jesper-friis has expressed he wants to have doc-strings for tests to explicitly clarify where the fixtures are coming from.

tests/ontopy_tests/test_graph2.py
```diff
@@ -1,14 +1,43 @@
...
+        # Add a relation that does not have a default style in OntoGraph
+        class hasSpecialRelation(owlready2.ObjectProperty):
+            "New special relation"
```
CasperWA, line 19:
```suggestion
            """New special relation"""
```

tests/ontopy_tests/test_graph2.py
```diff
@@ -1,14 +1,43 @@
...
+            domain = list(testonto.classes())
+            range = list(testonto.classes())
 
+        class NewSpecialClass(owlready2.Thing):
+            "New class"
```
CasperWA, line 24:
```suggestion
            """New class"""
```

tests/ontopy_tests/test_graph2.py
```diff
@@ -1,14 +1,43 @@
...
+        relations="all",
+        addnodes=True,
+        edgelabels=None,
+    )
+    graph.add_legend()
+    graph.save(tmpdir / "testonto.png")
+
+
+def test_emmo_graphs(emmo: "Ontology", tmpdir: "Path") -> None:
     graph = OntoGraph(
```
CasperWA, line 41: While you didn't change or update this test explicitly, I see this is also missing a doc-string similar to the I suggested above.

CasperWA (changes requested):
I don't see any content issues here.

Also, I'm guessing the change in `graph.py` is what is being tested in `test_graph2.py`? :) So I think the comment may be ignored. However, it speaks to a variable-naming issue.

tests/ontopy_tests/test_graph2.py
francescalb: I want to eventually add that the pytest fails upong generation of warnings. Sometimes the warnings are desired, therefore this filterwarningstatement. I am adding explicitly an object property that does not have a default style, and I want this to recognized. I guess I could also use pytest.warns(), but I think this is enough.

ontopy/utils.py
francescalb: As default? I guess this could be discussed with Jesper. He introduced this recently and I disagree :) Also, right now it doesn't work. I'll make an issue about it.

francescalb:
> 

changed the name.

ontopy/graph.py
CasperWA: Okay.

This seems quite confusing to me altogether. But if it's easily readable and understandable in the code (even when coming back to this after some weeks), I don't see an issue. Otherwise, it might be time to think about how to improve this to make it more legible.

tests/ontopy_tests/test_graph2.py
CasperWA: Filterwarnings _ignores_ warnings that occur, while `pytest.warns()` checks warnings _do_ occur. So, as I read the test you're _ignoring_ a specific warning in the output only. I'm pretty sure the warning is still emitted, just ignored in the printouts. If you want to test a specific warning is emitted, you should use `pytest.warns()`.

ontopy/graph.py
```diff
@@ -691,6 +691,27 @@ def get_node_attrs(self, name, nodeattrs, attrs):
         kwargs.update(attrs)
         return kwargs
 
+    def _relation_styles(self, entity, relations, rels, name):
+        """Helper function that returns the styles of the relations
+        to be used.
+        """
```
CasperWA, line 697: This is a great method abstraction. It also provides an excellent opportunity to describe the differences between `relations` and `rels`. Or indeed, just describe all of the arguments in the doc-string :)

CasperWA (changes requested):

ontopy/utils.py
francescalb: Using lowerlabel introduced an error. So we need to keep label. I have added that lowerlabel makes ontograph stypes fail in the issue.

tests/ontopy_tests/test_graph2.py
francescalb: Yes, in this specific example it ignores a very specific warning. If we at some point add (which I intend to) pytest failing on warnings, then it will fail if any other warning is emitted.

ontopy/graph.py
francescalb: I hope this is OK now.

tests/ontopy_tests/test_graph2.py
francescalb: Anyhow, changed it to testing for that exact warning now, so hope it is OK.

ontopy/graph.py
```diff
@@ -691,9 +692,43 @@ def get_node_attrs(self, name, nodeattrs, attrs):
...
+        self, entity: ThingClass, relations: dict, rels: set
+    ) -> dict:
+        """Helper function that returns the styles of the relations
+        to be used.
+
+        Parameters:
+            entity: the entity of the parent relation
+            relations: relations with default styles
+            rels: relations to be considered that have default styles,
+                either for the prefLab or one of the altLabels
```
CasperWA, line 705:
```suggestion
                either for the prefLabel or one of the altLabels
```

CasperWA (approved):
Small suggested change for a typo. Then all good to go 👍

PR #569, opened 2023-03-13
ontograph: colour updates, examples, bugfix

francescalb:
# Description:
* Updated colours so that all object properties currently in use in EMMO get a default colour (many as subproperties of a  property with default colour).

* The parents option in ontograph had a bug that is fixed.

* Added also som examples for ontograph.

* Updated pytest on writing emmographs to fail on warnings, in order to pick up if graph generation of EMMO suddensly starts haveing issues.

* Documentation stated that ontograph would guess format from file extension. This was not implemented, so moved the guess_format from ontodoc to utils and made use of it in ontograph.

## Type of change:

- [x] Bug fix.
- [ ] New feature.
- [x] Documentation update.

## Checklist:

### DIFF
--- a/ontopy/utils.py
+++ b/ontopy/utils.py
@@ -724,3 +724,14 @@ def normalise_url(url):
     components = list(splitted)
     components[2] = os.path.normpath(splitted.path)
     return urllib.parse.urlunsplit(components)
+
+
+def get_format(outfile: str, default: str, fmt: str = None):
+    """Infer format from outfile and format."""
+    if fmt is None:
+        fmt = os.path.splitext(outfile)[1]
+    if not fmt:
+        fmt = default
+    if fmt.startswith("."):
+        fmt = fmt[1:]
+    return fmt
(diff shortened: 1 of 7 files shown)


ontopy/utils.py
```diff
@@ -724,3 +724,14 @@ def normalise_url(url):
...
+
+def get_format(outfile: str, default: str, fmt: str = None):
+    """Infer format from outfile and format."""
+    if fmt is None:
+        fmt = os.path.splitext(outfile)[1]
+    if not fmt:
+        fmt = default
+    if fmt.startswith("."):
+        fmt = fmt[1:]
+    return fmt
```
jesper-friis, line 737: Alternatively
```suggestion
    return fmt.lstrip(".")
```

jesper-friis (approved):
Looks good. I approve, but please take a look at the comments. 
Especially the print statements should be removed.

jesper-friis (approved):
I like the bold legend.

PR #572, opened 2023-03-16
get_ancestors and get_descendants have the same arguments.

francescalb:
They now have the same arguments:
classes, generations and common.

Tests for get_ancestors have been added.

# Description:

## Type of change:

- [ ] Bug fix.
- [x] New feature.
- [ ] Documentation update.

## Checklist:

### DIFF
--- a/ontopy/graph.py
+++ b/ontopy/graph.py
@@ -400,15 +400,15 @@ class OntoGraph:  # pylint: disable=too-many-instance-attributes
             nodeattrs=nodeattrs,
             **attrs,
         )
-        common_ancestors = False
+        closest_ancestors = False
         ancestor_generations = None
-        if include_parents in ("common", "closest"):
-            common_ancestors = True
+        if include_parents == "closest":
+            closest_ancestors = True
         elif isinstance(include_parents, int):
             ancestor_generations = include_parents
         parents = self.ontology.get_ancestors(
             classes,
-            common=common_ancestors,
+            closest=closest_ancestors,
             generations=ancestor_generations,
             strict=True,
         )
--- a/ontopy/graph.py
+++ b/ontopy/graph.py
@@ -400,9 +400,17 @@ class OntoGraph:  # pylint: disable=too-many-instance-attributes
             nodeattrs=nodeattrs,
             **attrs,
         )
-
+        common_ancestors = False
+        ancestor_generations = None
+        if include_parents in ("common", "closest"):
+            common_ancestors = True
+        elif isinstance(include_parents, int):
+            ancestor_generations = include_parents
         parents = self.ontology.get_ancestors(
-            classes, include=include_parents, strict=True
+            classes,
+            common=common_ancestors,
+            generations=ancestor_generations,
+            strict=True,
         )
         if parents:
             for parent in parents:
--- a/ontopy/graph.py
+++ b/ontopy/graph.py
@@ -371,7 +371,7 @@ class OntoGraph:  # pylint: disable=too-many-instance-attributes
         also included."""
         if leaves is None:
             leaves = ()
-
+        print(leaves)
         classes = self.ontology.get_branch(
             root=root,
             leaves=leaves,
--- a/ontopy/graph.py
+++ b/ontopy/graph.py
@@ -371,7 +371,6 @@ class OntoGraph:  # pylint: disable=too-many-instance-attributes
         also included."""
         if leaves is None:
             leaves = ()
-        print(leaves)
         classes = self.ontology.get_branch(
             root=root,
             leaves=leaves,
--- a/ontopy/graph.py
+++ b/ontopy/graph.py
@@ -810,7 +810,7 @@ class OntoGraph:  # pylint: disable=too-many-instance-attributes
         label1 = [table]
         label2 = [table]
         for index, relation in enumerate(relations):
-            if relation in rels:
+            if (relation in rels) or (relation == "isA"):
                 label1.append(
                     f'<tr><td align="right" '
                     f'port="i{index}"><b>{relation}</b></td></tr>'
--- a/ontopy/ontology.py
+++ b/ontopy/ontology.py
@@ -4,7 +4,7 @@
 If desirable some of these additions may be moved back into owlready2.
 """
 # pylint: disable=too-many-lines,fixme,arguments-differ,protected-access
-from typing import TYPE_CHECKING, Optional, Union, Sequence
+from typing import TYPE_CHECKING, Optional, Union
 import os
 import itertools
 import inspect
@@ -1468,18 +1468,25 @@ class Ontology(owlready2.Ontology):  # pylint: disable=too-many-public-methods
             "A closest common ancestor should always exist !"
         )
 
-    def get_ancestors(self, classes, include="all", strict=True):
+    def get_ancestors(
+        self,
+        classes: "Union[List, ThingClass]",
+        common: bool = False,
+        generations: int = None,
+        strict: bool = True,
+    ) -> set:
         """Return ancestors of all classes in `classes`.
-        classes to be provided as list
-
-        The values of `include` may be:
-          - None: ignore this argument
-          - "all": Include all ancestors.
-          - "closest": Include all ancestors up to the closest common
-            ancestor of all classes.
-          - int: Include this number of ancestor levels.  Here `include`
-            may be an integer or a string that can be converted to int.
+        Args:
+            classes: class(es) for which ancestors are desired.
+            common: whether to only return the closest common ancestor.
+            generations: Include this number of generations, default is all.
+            strict: only return real ancestors if True.
+        Returns:
+            A set of ancestors for given number of generations.
         """
+        if not isinstance(classes, Iterable):
+            classes = [classes]
+
         ancestors = set()
         if not classes:
             return ancestors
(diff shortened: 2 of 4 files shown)


ontopy/ontology.py
```diff
@@ -1468,18 +1468,25 @@ def closest_common_ancestor(*classes):
...
-        The values of `include` may be:
-          - None: ignore this argument
-          - "all": Include all ancestors.
-          - "closest": Include all ancestors up to the closest common
-            ancestor of all classes.
-          - int: Include this number of ancestor levels.  Here `include`
-            may be an integer or a string that can be converted to int.
+        Args:
+            classes: class(es) for which ancestors are desired.
+            common: whether to only return the closest common ancestor.
```
jesper-friis, line 1481:
```suggestion
            common: whether to only return the closest common ancestor of `classes`.
```

ontopy/ontology.py
```diff
@@ -1468,18 +1468,25 @@ def closest_common_ancestor(*classes):
...
+            generations: Include this number of generations, default is all.
+            strict: only return real ancestors if True.
```
jesper-friis, line 1483:
```suggestion
            strict: only return real ancestors if True, i.e. not include self.
```

ontopy/ontology.py
francescalb: There already is a closest_common_ancestor, so no reason to have it here. 

The `common` is equivalent to the include='closest' previously, which would only return the ancestors up to the closest common ancestor. Maybe closest is a better argument?

The reason for having 'common' was that we previously discussed making the to get_ancestors and get_descendants more harmonized. In get_descendants, however, common only returns common descendants. 

maybe we should have the argument `closest` in get_ancestors and `common` in get_descendants and accept that the two functions will not have the same arguments?

jesper-friis: Thank, now I understand. 

Yes, I think that common is misleading, since the intention is to return the not-common ancestors. I think we should also be more clear in the documentation of the `common` argument.

By the way, `strict` seems to be ignored when `common=True`. The expected behaviour would be to exclude the common ancestor when `strict=True` (i.e. CompositeParticle in the case of `emmo.get_ancestors((onto.Atom, onto.Electron), common=True, strict=True)`.

francescalb: No, strict is only to exlude the classes themselves, i.e. only have true ancestors. __

jesper-friis: You are right. We should not give more meanings.

ontopy/ontology.py
francescalb: - rewritten

ontopy/graph.py
```diff
@@ -400,9 +400,17 @@ def add_branch(  # pylint: disable=too-many-arguments,too-many-locals
             nodeattrs=nodeattrs,
             **attrs,
         )
-
+        closest_ancestors = False
+        ancestor_generations = None
+        if include_parents == "closest":
+            closest_ancestors = True
+        elif isinstance(include_parents, int):
+            ancestor_generations = include_parents
```
jesper-friis, line 408: Må testes - interessant info fra codecov

jesper-friis (approved):
Approved. I will leave it to you to decide whether to add a separate test for line 408

ontopy/graph.py
```diff
@@ -371,7 +371,7 @@ def add_branch(  # pylint: disable=too-many-arguments,too-many-locals
         also included."""
         if leaves is None:
             leaves = ()
-
+        print(leaves)
```
jesper-friis, line 374: Please remove

ontopy/ontology.py
```diff
@@ -1490,22 +1499,23 @@ def addancestors(entity, counter, subject):
...
                     addancestors(parent, counter - 1, subject)
 
-        if isinstance(include, str) and include.isdigit():
-            include = int(include)
-
-        if include == "all":
-            ancestors.update(*(_.ancestors() for _ in classes))
-        elif include == "closest":
-            closest = self.closest_common_ancestor(*classes)
+        if sum(map(bool, [closest, generations])) > 1:
```
jesper-friis, line 1502: Easier to read if you write
```python
        if closest and generations is not None:
```
Another benefit is that it handles the case of setting `generations=0` correctly.

It will be slightly more efficient (with no cost in readability) if you place this check within the if-statement on line 1506, like:
```python
if closest:
    if generations is not None:
        raise ValueError(...)
    closest_ancestor = ...
    ...
```

ontopy/ontology.py
```diff
@@ -1490,22 +1499,23 @@ def addancestors(entity, counter, subject):
...
+            raise ValueError(
+                "Only one of `generations` or `closest` may be specified."
+            )
+        if closest:
+            closest_ancestor = self.closest_common_ancestor(*classes)
             for cls in classes:
                 ancestors.update(
-                    _ for _ in cls.ancestors() if closest in _.ancestors()
+                    _
```
jesper-friis, line 1510: The anonymous variable name `_` should be reserved for cases where we explicitly are ignore a variable. For instance, if we have a list of triples and only are interested in looping over the subjects, it is ok to write:

```python
triples = [('s1', 'p1', 'o1'), ('s2', 'p2', 'o2')]
for s, _, _ in triples:
    # so something with `s`
```

Please note that `_` also has a special meaning in the Python interpreter - it is always assigned to the latest result. For example

```python
>>> 1 + 1
2
>>> _
2
```
Starting to use it as a variable can create confusion - for instance if someone wants to debug your code and paste your code lines into the interpreter, statement for statement, the value of `_` will be overwritten by the interpreter...

In this case I would use name `a` as a good, short and readable name for the temporary variable. That it is a single letter carries the semantics of it being a temporary variable.

ontopy/ontology.py
```diff
@@ -1490,22 +1499,23 @@ def addancestors(entity, counter, subject):
...
                 )
-        elif isinstance(include, int):
+        elif isinstance(generations, int):
             for entity in classes:
-                addancestors(entity, int(include), ancestors)
-        elif include not in (None, "None", "none", ""):
-            raise ValueError('include must be "all", "closest" or None')
+                addancestors(entity, generations, ancestors)
+        else:
+            ancestors.update(*(_.ancestors() for _ in classes))
```
jesper-friis, line 1518: As avove.  Here I would use `c` instead of `_`.

ontopy/ontology.py
```diff
@@ -1529,7 +1539,7 @@ def get_descendants(
             'generations' defaults to all.
         """
 
-        if not isinstance(classes, Sequence):
+        if not isinstance(classes, Iterable):
```
jesper-friis, line 1542: Well spotted!

tests/ontopy_tests/test_graph.py
```diff
@@ -77,7 +77,7 @@ class hasPartRenamed(owlready2.ObjectProperty):
     graph.save(tmpdir / "testonto.png")
 
 
-def test_emmo_graphs(emmo: "Ontology", tmpdir: "Path") -> None:
+def test_emmo_graphs(emmo: "Ontology", repo_dir: "Path") -> None:
```
jesper-friis, line 80: Please also rename the variable name in the documentation below accordingly

tests/ontopy_tests/test_graph.py
```diff
@@ -90,7 +90,7 @@ def test_emmo_graphs(emmo: "Ontology", tmpdir: "Path") -> None:
     from ontopy.graph import OntoGraph, plot_modules
     from ontopy import get_ontology
     import warnings
-
+    tmpdir=repo_dir / "tmp"
```
jesper-friis, line 93:
```suggestion
    tmpdir = repo_dir / "tmp"
```

tests/test_generation_search.py
```diff
@@ -80,3 +80,70 @@ def test_descendants(emmo: "Ontology", repo_dir: "Path") -> None:
     assert onto.get_descendants([onto.Tree, onto.NaturalDye], common=True) == {
         onto.Avocado
     }
+
+
+def test_ancestors(emmo: "Ontology", repo_dir: "Path") -> None:
```
jesper-friis, line 85: Nice test

jesper-friis (commented):
Added some comments

jesper-friis (commented):
Added some comments

ontopy/ontology.py
```diff
@@ -1468,18 +1468,27 @@ def closest_common_ancestor(*classes):
...
-        classes to be provided as list
-
-        The values of `include` may be:
-          - None: ignore this argument
-          - "all": Include all ancestors.
-          - "closest": Include all ancestors up to the closest common
-            ancestor of all classes.
-          - int: Include this number of ancestor levels.  Here `include`
-            may be an integer or a string that can be converted to int.
+        Args:
```
jesper-friis, line 1479:
```suggestion
        Arguments:
```

ontopy/ontology.py
francescalb: Tahnk you for really informative teaching!

tests/ontopy_tests/test_graph.py
francescalb: Well spotted. I just changed it to repo_dir temporarily to see the actual output.I will change back to tmpdir.

tests/ontopy_tests/test_graph.py
francescalb: same as  above

ontopy/ontology.py
francescalb: Why: I have kept the style of the file. It says Args in all, and not Arguments.

ontopy/graph.py
francescalb: Testet

ontopy/ontology.py
jesper-friis: Then we keep Args.

jesper-friis (approved):

PR #577, opened 2023-04-09
Added two additional exceptions to emmocheck

jesper-friis:
# Description:
Added the following two exceptions to emmocheck
- foaf.logo comes from foaf and does not have a prefLabel
- Accept 3DPrinting in the DOAM ontology as a valid prefLabel, even though it does not start with a letter

## Type of change:

- [ ] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

## Checklist:

### DIFF
--- a/emmopy/emmocheck.py
+++ b/emmopy/emmocheck.py
@@ -120,7 +120,6 @@ class TestSyntacticEMMOConventions(TestEMMOConventions):
                 "2-manifold",
                 "3-manifold",
                 "C++",
-                "3DPrinting",
             )
         )
         exceptions.update(self.get_config("test_class_label.exceptions", ()))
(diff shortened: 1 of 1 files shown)


francescalb (changes requested):
I do not think the correct way to go about exceptions to emmocheck for ontologies except emmo itself is to add disregarded labels in emmocheck at top level. Then one should be able to add a list of labels to disregard in emmocheck.

jesper-friis:
> I do not think the correct way to go about exceptions to emmocheck for ontologies except emmo itself is to add disregarded labels in emmocheck at top level. Then one should be able to add a list of labels to disregard in emmocheck.

I fully agree. But these concepts are in EMMO.

jesper-friis:
Correction. 3DPrinting is not in EMMO any more - so I have removed the exception for that.

francescalb (approved):

PR #582, opened 2023-04-11
Updated get_by_label() so that it now accepts label, name and full iri

### LINKED ISSUE #583: Allow space in labels
Motivation: make sure that ontopy supports ontologies that have spaces in their labels, like BFO.

The emmocheck tool will still make sure that EMMO-based ontologies have no space in their prefLabel.

jesper-friis:
# Description:
The main issue addressed in this PR is to update get_by_label() such that it allows access by name (what follows the hash in the IRI).

However, when we are at it, a few other things have been included:
- access by IRI
- documented how conflict in prefix in label and argument is handled
- allow space in labels
- added test for label, name and full IRI
- added test for dir listing on the ontology (due to codecov complaining)

Closes #583 

## Type of change:

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

## Checklist:

### DIFF
--- a/ontopy/ontology.py
+++ b/ontopy/ontology.py
@@ -187,10 +187,10 @@ class Ontology(owlready2.Ontology):  # pylint: disable=too-many-public-methods
     )
 
     # Other settings
-    _colon_in_name = False
-    colon_in_name = property(
-        fget=lambda self: self._colon_in_name,
-        fset=lambda self, v: setattr(self, "_colon_in_name", bool(v)),
+    _colon_in_label = False
+    colon_in_label = property(
+        fget=lambda self: self._colon_in_label,
+        fset=lambda self, v: setattr(self, "_colon_in_label", bool(v)),
         doc="Whether to accept colon in name-part of IRI.  "
         "If true, the name cannot be prefixed.",
     )
@@ -270,7 +270,7 @@ class Ontology(owlready2.Ontology):  # pylint: disable=too-many-public-methods
         label_annotations: str = None,
         prefix: str = None,
         imported: bool = True,
-        colon_in_name: bool = None,
+        colon_in_label: bool = None,
     ):
         """Returns entity with label annotation `label`.
 
@@ -286,9 +286,9 @@ class Ontology(owlready2.Ontology):  # pylint: disable=too-many-public-methods
                (#) stripped off).  The search for a matching label will be
                limited to this namespace.
            imported: Whether to also look for `label` in imported ontologies.
-           colon_in_name: Whether to accept colon (:) in name-part of IRI.
-               Defaults to the `colon_in_name` property of `self`.
-               A true value cannot be combined with `prefix`.
+           colon_in_label: Whether to accept colon (:) in a label or name-part
+               of IRI.  Defaults to the `colon_in_label` property of `self`.
+               Setting this true cannot be combined with `prefix`.
 
         If several entities have the same label, only the one which is
         found first is returned.Use get_by_label_all() to get all matches.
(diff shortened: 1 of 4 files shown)


ontopy/ontology.py
```diff
@@ -283,39 +283,45 @@ def get_by_label(
...
                 f"Invalid label definition, must be a string: {label!r}"
             )
+
+        # Comment to reviewer:
+        # Do we really want to disallow space in label?
+        # I think we wanted ontopy to work with any ontology, also ontologies
+        # having space in their labels.
+        # Rather than enforcing EMMO conventions here, I think the user should
+        # run emmocheck on the ontology - it will alert about labels with
+        # spaces
```
francescalb, line 298: I agree, please add this as an issue.

ontopy/ontology.py
```diff
@@ -283,39 +283,45 @@ def get_by_label(
         The current implementation also supports "*" as a wildcard
         matching any number of characters. This may change in the future.
         """
-        # pylint: disable=too-many-arguments,too-many-branches
+        # pylint: disable=too-many-arguments,too-many-branches,invalid-name
```
francescalb, line 286: Please use 'valid names' instead.

ontopy/ontology.py
```diff
@@ -283,39 +283,45 @@ def get_by_label(
...
             for iri in DEFAULT_LABEL_ANNOTATIONS:
                 try:
                     self.add_label_annotation(iri)
                 except ValueError:
                     pass
 
+        # Comment to reviewer:
+        # Since providing the prefix as argument is more explicit, I think
+        # that should take precedence.  Changed it below...
         splitlabel = label.split(":", 1)
```
francescalb, line 314: Ok. Please remove this comment and add it in the doc-string.

francescalb (changes requested):
Please consider the requested changes and I can review again,

ontopy/ontology.py
jesper-friis: Issue added. Ok to fix and close it in this PR?
We just have to remove the next 4 lines.

ontopy/ontology.py
```diff
@@ -283,39 +283,45 @@ def get_by_label(
...
-        if len(splitlabel) > 2:
-            raise ValueError(
-                f"Invalid label definition, {label!r}"
-                " contains more than one ':' ."
-                "The string before ':' indicates the prefix. "
-                "The string after ':' indicates the label."
-            )
```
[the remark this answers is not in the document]
jesper-friis: I removed it because `len(splitlabel)` is always equal to or less than 2, hence raise will never be called.
This is because of the second argument to `split()` on line 302 ensuring that the label is only splittet once.

But simply removing this check may be the wrong medicine. 

The issue is little tricky. On one hand we don't want to exclude using ontopy with ontologies having colon in their labels. On the other hand, it is difficult to distinguish between a colon indicating a prefix and a colon that is part of the label. 

Should we add a new boolean argument `colon_ok` together with a corresponding ontology-wise setting, about whether to labels with colon are acceptable (in which case label prefixes are turned off). Or is it too much hassle for a corner case?

ontopy/ontology.py
francescalb: OK

ontopy/ontology.py
francescalb: My point was not to introduce pylint exceptions.

ontopy/ontology.py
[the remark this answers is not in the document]
francescalb: I think that is a good idea.

ontopy/ontology.py
jesper-friis: But I thought that we all agree that s, p, o are valid names in triples.

francescalb: OK, but not for the a in lines 330 to 332

jesper-friis: OK, I added some extra characters to those `a`'s. But I really don't think it improves anything - just makes the lines longer...

ontopy/ontology.py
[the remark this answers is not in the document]
jesper-friis: Done

francescalb (approved):
Looks good. Please add a colon in name in the test, and then it should be good to go.

PR #600, opened 2023-05-05
Add links to the original FaCT++ repo, GitHub profiles, etc.

blokhin:
# Description:
Add more info about FaCT++ reasoner

## Type of change:
- [ x] Documentation update.

### DIFF
--- a/ontopy/factpluspluswrapper/README.md
+++ b/ontopy/factpluspluswrapper/README.md
@@ -1,8 +1,8 @@
-Python interface to the FaCT++ Reasoner.
+Python interface to the [FaCT++ Reasoner](https://bitbucket.org/dtsarkov/factplusplus).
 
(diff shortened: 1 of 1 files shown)


jesper-friis:
Thank you Evgeny. It would also be nice to have a link to your FaCT++ repo for later reference.

jesper-friis (approved):

francescalb (approved):

blokhin:
@jesper-friis here it is https://github.com/tilde-lab/pyfactxx its obvious advantage over your current implementation is that the C++ kernel is directly used, and there is no need to invoke (and require) java APIs. I expect the replacement should not be complicated, since `pyfactxx` also provides the `rdflib` store.

jesper-friis:
Thank you @blokhin. I added an issue on this: https://github.com/emmo-repo/EMMOntoPy/issues/603

PR #610, opened 2023-05-30
Change ontology.new_entity to also allow adding properties

francescalb:
# Description
new_entity adds a ThingClass as default  as before, but it is now possble to add annotation/object/dataproperties as well.

## Type of change

- [ ] Bug fix.
- [x] New feature.
- [x] Documentation update.
- [x] Test update.

## Checklist

### DIFF
--- a/ontopy/ontology.py
+++ b/ontopy/ontology.py
@@ -1649,17 +1649,22 @@ class Ontology(owlready2.Ontology):  # pylint: disable=too-many-public-methods
                 AnnotationPropertyClass,
             ]
         ] = "class",
-    ) -> ThingClass:
+    ) -> Union[
+        ThingClass,
+        ObjectPropertyClass,
+        DataPropertyClass,
+        AnnotationPropertyClass,
+    ]:
         """Create and return new entity
 
         Args:
             name: name of the entity
             parent: parent(s) of the entity
             entitytype: type of the entity,
-            default is 'class/ThingClass'. Other options
-            are data_property, object_property,
-            annotation_property or ObjectPropertyClass,
-            DataPropertyClass and AnnotationProperty classes.
+                default is 'class/ThingClass'. Other options
+                are data_property, object_property,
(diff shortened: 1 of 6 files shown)


ontopy/utils.py
```diff
@@ -246,17 +248,20 @@ def fmt(entity):
...
-    if issubclass(expr, (bool, int, float, str)):
-        return fmt(expr.__class__.__name__)
-    if issubclass(expr, datetime.date):
-        return "date"
-    if issubclass(expr, datetime.time):
-        return "datetime"
-    if issubclass(expr, datetime.datetime):
-        return "datetime"
+    if inspect.isclass(expr):
+        print("expr", expr, type(expr))
```
francescalb, line 256:
```suggestion
```

ontopy/utils.py
```diff
@@ -136,6 +137,7 @@ def asstring(  # pylint: disable=too-many-return-statements,too-many-branches,to
     """
     if ontology is None:
         ontology = expr.ontology
+    print("exprtio", expr)
```
francescalb, line 140:
```suggestion
```

ontopy/ontology.py
```diff
@@ -1629,32 +1631,150 @@ def get_wu_palmer_measure(self, cls1, cls2):
...
-                raise ThingClassDefinitionError(
+            if not isinstance(thing, parenttype):
+                raise EntityClassDefinitionError(
                     f"Error in parent definition: "
-                    f"'{thing}' is not an owlready2.ThingClass."
+                    f"'{thing}' is not an {parenttype}."
                 )
 
         with self:
             entity = types.new_class(name, parents)
```
alfredoisg, line 1715: I may be lacking context but, isn't this missing some kind of reference to `entitytype`?

francescalb: Parenttype is set above depending on entitytype. It is done like this so that you can either choose to give entitytype as a str (which then needs to be converted) or give it as a Python class directly.

PR #618, opened 2023-06-07
Fixed failing tests in test_patch.py

jesper-friis:
# Description
Due to resent updates in EMMO, several of the tests in test_patch.py were failing.

We should consider to test against our test ontology instead of EMMO to avoid such failures.

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [x] Test update.

## Checklist

### DIFF
--- a/tests/ontopy_tests/test_patch.py
+++ b/tests/ontopy_tests/test_patch.py
@@ -52,8 +52,9 @@ assert emmo.Holistic.is_defined == True
 # Comment out these tests for now because Owlready2 automatically converts
 # `Inverse(emmo.hasPart)` to `emmo.isPartOf`.
 #
-# Also, decide whether we really want to keep get_indirect_is_a() - it
-# very much dublicates what ancestors() already does.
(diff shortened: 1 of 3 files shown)


tests/ontopy_tests/test_patch.py
```diff
@@ -44,37 +41,36 @@
 assert emmo.Atom["altLabel"] == ["ChemicalElement"]
 
 
+assert emmo.Atom.is_defined == False
```
francescalb, line 44: Is_defined should be tested as part of test_ontopy as it is defined there.

francescalb (changes requested):
left one comment

tests/ontopy_tests/test_patch.py
jesper-friis: We are testing the `is_defined` property of Atom, which is patched. The implementation in ontology is deprecated.

francescalb (commented):
Very good, but please add an issue about the tests that are going to be added back.

PR #623, opened 2023-06-09
default_annotations no longer forced in get_by_label

francescalb:
* label_annotations are not set in get by label
* get_by_label and get_by_label_all also look up names
* get_by_label_all return set instead of list
* setting label annotations must be done explicitly in excelparser

# Description

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [x] Test update.

## Checklist


ontopy/ontology.py
```diff
@@ -343,12 +344,13 @@ def get_by_label(
                 label_annotations=label_annotations,
                 prefix=prefix,
             )
-            if len(entitylist) > 0:
-                return entitylist[0]
+            if len(entitylist) == 1:
+                return next(iter(entitylist))
```
jesper-friis, line 348: `entitylist` er ikke lenger en liste, men et sett. Burde vi endre navn på den?

Vi ønsker å sjekke om det finnes entitylist, ikke om den kun har et element:

```suggestion
            if entitylist:
                return entitylist.pop()
```

ontopy/ontology.py
```diff
@@ -343,12 +344,13 @@ def get_by_label(
...
 
             raise NoSuchLabelError(
-                f"No label annotations matches {label!r} with prefix "
-                f"{prefix!r}"
+                f"Either no label annotations matches for {label!r} "
```
jesper-friis, line 351: Maybe better to separate the two cases with `NoSuchLabelError` (entitylist is empty) and `AmbigiousLabelError` (entitylist has more than one element).

ontopy/ontology.py
```diff
@@ -409,26 +416,38 @@ def get_by_label_all(self, label, label_annotations=None, prefix=None):
...
-        else:
+        if label_annotations:
             annotations = (
                 _.name if hasattr(_, "storid") else _ for _ in label_annotations
             )
-        entity = self.world.search(**{next(annotations): label})
-        for key in annotations:
-            entity.extend(self.world.search(**{key: label}))
+        elif self._label_annotations:
+            annotations = (_.name for _ in self.label_annotations)
```
jesper-friis, line 424: In general it is bad practice to use `_` as a variable name. It is fine to use as a placeholder for an unused variable, but not for something we are using. Reasons: (1) it is commonly used indicate an unused variable, let not give it more semantics and (2) it is assigned to the last result in many interpreters, which can lead to surprises for people that test code snippets by copy-paste into an interpreter. 
```suggestion
            annotations = (la.name for la in self.label_annotations)
```

ontopy/ontology.py
```diff
@@ -409,26 +416,38 @@ def get_by_label_all(self, label, label_annotations=None, prefix=None):
...
+
+        else:
+            annotations = None
+        entities = set()
+        if annotations:
+            # entity = self.world.search(**{next(annotations): label})
```
jesper-friis, line 430: Can be removed

tests/ontopy_tests/test_new_entity.py
```diff
@@ -21,11 +21,12 @@ def test_new_entity(testonto: "Ontology") -> None:
 
     # Test that new entity is found by both version of get_by_label
     assert testonto.get_by_label("FantasyClass") == testonto.FantasyClass
-    assert testonto.get_by_label_all("FantasyClass") == [testonto.FantasyClass]
+    print(testonto.get_by_label_all("*"))
```
jesper-friis, line 24: Do you want to keep this print statement?

jesper-friis (commented):
Added some comments, but it looks good.

jesper-friis (approved):

PR #639, opened 2023-08-15
Update to comply with owlready2>0.41

francescalb:
# Description
Olwready2 has changed so that when asking for an attribute it returns its name as a locstr (e.g. locstr('Atom', 'en')) instead of jsut the string (e.g. 'Atom'). This enables differentiating between languages. 
locstr inherits from str so mostly it does not change much, except that the output looks different. 

A setassert function was made to help testing. 

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [x] Test update.
 

## Checklist


ontopy/excelparser.py
```diff
@@ -562,7 +559,7 @@ def get_metadata_from_dataframe(  # pylint: disable=too-many-locals,too-many-bra
             )
         except AttributeError:
             pass
-
+    print("************", onto.elucidation, onto.Atom.elucidation)
```
jesper-friis, line 562: Remove this print statement

requirements.txt
jesper-friis, line 1: In principle unrelated to this PR, but I think that we should change `rdflib <7`. In v8 rdflib adds an `overwrite` option that it supplies to some callbacks. Currently the Owlready2 triplestore backend does not accept this option, resulting in an error.

tests/conftest.py
```diff
@@ -6,6 +6,10 @@
 
 if TYPE_CHECKING:
     from ontopy.ontology import Ontology
+    from typing import Sequence
+import sys, os
+
+sys.path.append(os.path.join(os.path.dirname(__file__), "utilities"))
```
jesper-friis, line 12: Would be nice with a comment explaining the purpose of this?

tests/ontopy_tests/test_patch.py
```diff
@@ -1,48 +1,91 @@
...
 import pytest
 
 from ontopy import get_ontology
 
 from owlready2 import owl, Inverse
 
+from utilities import setassert
+
+
+def test_get_by_label_onto(emmo: "Ontology") -> None:
```
jesper-friis, line 14: What version of EMMO does the `emmo` argument correspond to? Does it correspond to a stable tag or will it change with time?

tests/ontopy_tests/test_prefix.py
```diff
@@ -37,6 +38,12 @@ def test_prefix(testonto: "Ontology", emmo: "Ontology") -> None:
         "TestClass", prefix="models"
     )
 
+    with pytest.raises(ValueError):
+        assert testonto.get_by_label_all(" ")
```
jesper-friis, line 42: The `assert` statement is not used (you get the ValueError exception before the assertment is executed). Easier to understand the test if you remove it.

tests/ontopy_tests/test_prefix.py
```diff
@@ -37,6 +38,12 @@ def test_prefix(testonto: "Ontology", emmo: "Ontology") -> None:
...
+
+    with pytest.raises(TypeError):
+        assert testonto.get_by_label(1)
```
jesper-friis, line 45: As above, the `assert` statement can be removed.

requirements.txt
francescalb: Do you have an example of this?

tests/conftest.py
francescalb: Done

tests/genfigs/TestClass.png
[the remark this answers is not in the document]
francescalb: Nope, removed

jesper-friis (commented):
Nice. Added some comments

requirements.txt
jesper-friis: Hmm, maybe it was something that depends on rdflib that caused the issue had this morning when working with emmocheck. It works now when I selectively upgrade rdflib from v6.0.0 to v7.0.0. 

Lets not spend more time on this.

ontopy/ontology.py
```diff
@@ -904,7 +904,7 @@ def save(
             )
 
         revmap = {value: key for key, value in FMAP.items()}
-
+        print(filename)
```
jesper-friis, line 907: A leftover debugging print statement?

tests/ontopy_tests/test_patch.py
jesper-friis: Ok. Since the pytest fixtures are so opaque, it might be good to add a comment saying this (since we might have to update the tests when get_ontology() is updated to return a newer version of EMMO).

francescalb: I do not agree that they are opaque, but to make it more readily clear I added a comment in the emmo fixture on conftest.

jesper-friis (approved):

PR #641, opened 2023-08-19
Simplified implementation of label_annotations.

jesper-friis:
# Description
Simplified the implementation of `label_annotations`. The `label_annotations` property is now a list of full IRI's to label annotations that will be used by `get_by_label()` given that the IRI is in the world of the ontology. 

Hence, if the ontology doesn't define `skos:prefLabel` it will not be used by `get_by_label()`  even though that "http://www.w3.org/2004/02/skos/core#prefLabel" may be included in `onto.label_annotations`.

I think this make things behave as expected in common cases without the user having to fickle with `label_annotations`.

The label annotations that are currently used by `get_by_label()` can be listed with 

    onto._to_entities(onto.label_annotations)

I also deprecated some functions that I only think increases the complexity without providing any real value.

See also comments IN CAPITALS. They should be removed before merging to master.

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [ ] Test update.

## Checklist

### DIFF
--- a/ontopy/ontology.py
+++ b/ontopy/ontology.py
@@ -97,7 +97,7 @@ class World(owlready2.World):
                   emmo-inferred and emmo-development will be the same.
             OntologyClass: If given and `base_iri` doesn't correspond
                 to an existing ontology, a new ontology is created of
-                this Ontology subclass.
+                this Ontology subclass.  Defaults to `ontopy.Ontology`.
             label_annotations: Sequence of label IRIs used for accessing
                 entities in the ontology given that they are in the ontology.
                 Label IRIs not in the ontology will simply be ignored.
@@ -305,7 +305,7 @@ class Ontology(owlready2.Ontology):  # pylint: disable=too-many-public-methods
     ):
         """Returns entity with label annotation `label`.
 
-        Args:
+        Arguments:
            label: label so serach for.
                May be written as 'label' or 'prefix:label'.
                get_by_label('prefix:label') ==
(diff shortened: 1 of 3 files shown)


ontopy/ontology.py
```diff
@@ -76,18 +76,32 @@ def __init__(self, *args, **kwargs):
...
+                - file name (possible excluding final .owl or .ttl)
+                - "emmo": load latest version of asserted EMMO
+                - "emmo-inferred": load latest version of inferred EMMO
+                  (default)
+                - "emmo-development": load latest inferred development
+                  version of EMMO. Until first stable release
+                  emmo-inferred and emmo-development will be the same.
+            OntologyClass: If given and `base_iri` doesn't correspond
+                to an existing ontology, a new ontology is created of
+                this Ontology subclass.
```
francescalb, line 100: What is the default?

jesper-friis: The default for OntologyClass is ontopy.Ontology

francescalb: Kind of, but as it is written now it is None, and if it is None it is then made to ontopy.Ontology. I don't really care how it is done, but at least it should be clear in the documentation what the default is (and now it seems like none).

ontopy/ontology.py
```diff
@@ -424,20 +428,28 @@ def get_by_label_all(self, label, label_annotations=None, prefix=None):
...
+
+        # ALTERNATIVE IMPLEMENTATION THAT DOESN'T INVOLVE search().
+        # IT IS FASTER AND MIGHT AVOID SURPRICES.  SHOULD WE ADD AN
+        # OPTION TO CHOOSE BETWEEN SEARCH() AND THIS IMPLEMENTATION?
+        # entities = set()
+        # for storid in self._to_storids(label_annotations):
+        #    entities.update(
+        #        self.world[self._unabbreviate(s)]
+        #        for s, _, _, _ in get_triples(None, storid, label, None)
+        #    )
```
francescalb, line 452: Why is it faster? If the outcome is exactly the same in both versions I do not see the reason to have two ways of doing the same thing. Then  think we should go for the fastest solution. Did you time it?

ontopy/ontology.py
```diff
@@ -1746,12 +1798,14 @@ def new_entity(
 
         with self:
             entity = types.new_class(name, parents)
+
+            # AREN'T WE LITTLE TOO SMART HERE?
```
francescalb, line 1802: Why? The idea is that if we make a new entity it is given prefLabel immediately. This can be changed of course afterwards, but generally I think this would be normal behaviour, no?

tests/test_get_by_label.py
```diff
@@ -4,25 +4,31 @@
...
     """Test that label annotations are added correctly if they are not added before
     using get_by_label
     """
     import owlready2
 
     testonto = get_ontology("http://domain_ontology/new_ontology")
     testonto.new_entity("Class", owlready2.Thing)
-    assert testonto._label_annotations == None
+
+    # THIS IS NOT NONE ANYMORE
```
francescalb, line 16: Can we then change it to what it is?

tests/test_get_by_label.py
```diff
@@ -33,29 +39,27 @@ def test_get_by_label_all_onto() -> None:
...
     )
     testonto.Class.SpecialAnnotation.append("This is a comment")
-    testonto.set_default_label_annotations()
 
     testonto.new_entity("Klasse", testonto.Class)
 
-    assert testonto.Klasse.prefLabel == ["Klasse"]
-
-    testonto.Klasse.altLabel = "Class2"
+    # THESE NOW FAILS, SINCE prefLabel AND altLabel ARE NOT DEFINED in testonto
```
francescalb, line 50: But what do we use instead if we make an ontology from scratch and then want to add the default label annotations? That is what this test was checking

ontopy/ontology.py
jesper-friis: Ok, documented the default.

ontopy/ontology.py
jesper-friis: Added exact_match argument as you suggested.

ontopy/ontology.py
jesper-friis: updated as agreed

francescalb (approved):
Approved - this is a PR into another devlopemnt branch. The docstring error can be fixed there.

PR #671, opened 2023-11-03
HermiT is default reasoner.

francescalb:
Ontoconvert used to accept only FaCT++ because
owlready2.reasoning._apply_reasoning_results() was broken. It added the inferred
rdfs:subClassOf relations as rdf:type relations.

# Description

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [x] Test update.

## Checklist


tests/ontopy_tests/test_reasoner.py
jesper-friis, line 1: Possible improvement of this test:

Does testonto contain enough semantics for the reasoners to work with, such that they can add additional relations?
If not, one could consider to add an object property to testonto with a domain and range to give the reasoner something to work with.
If/when the reasoners add something, that should be tested for.

tools/ontograph
```diff
@@ -100,12 +100,11 @@ def main(
...
         "--reasoner",
         nargs="?",
-        const="FaCT++",
+        const="HermiT",
         choices=["FaCT++", "HermiT", "Pellet"],
         help=(
-            'Run given reasoner on the ontology. Valid reasoners are "FaCT++" '
-            '(default), "HermiT" and "Pellet". Note: FaCT++ is preferred with '
-            "EMMO."
+            'Run given reasoner on the ontology. Valid reasoners are "HermiT '
```
jesper-friis, line 106:
```suggestion
            'Run given reasoner on the ontology. Valid reasoners are "HermiT" '
```

tests/ontopy_tests/test_reasoner.py
francescalb: Done

PR #677, opened 2023-11-21
Skip checking dimensional units for domain ontologies that doesn't load this class.

jesper-friis:
# Description
Do not check SI dimensional units if the checked domain ontology does not load that class.

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [ ] Test update.

## Checklist

### DIFF
--- a/emmopy/emmocheck.py
+++ b/emmopy/emmocheck.py
@@ -405,6 +405,11 @@ class TestFunctionalEMMOConventions(TestEMMOConventions):
 
     def test_dimensional_unit(self):
         """Check correct syntax of dimension string of dimensional units."""
+
+        # This test requires that the ontology has imported SIDimensionalUnit
+        if "SIDimensionalUnit" not in self.onto:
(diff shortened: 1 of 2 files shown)


francescalb (changes requested):
In PR#658 test_emmocheck_module was added, but it is not actually testing anything when run with pytest. So apart tests/tools/test_emmocheck there i nothing that tests emmocheck here. The tool only checks that it runs, but nothing more (it passes as test even if the emmocheck fails) 

Please update the test so that it is actually run and then that it fails when emmocheck fails.

jesper-friis:
> Please update the test so that it is actually run and then that it fails when emmocheck fails.

This PR just skip testing dimensionalities for ontologies that doesn't import the relevant modules. Isn't adding general tests for emmocheck another issue?

francescalb (approved):

francescalb (approved):
Appoved, but make an issue about creating a test.

PR #678, opened 2023-11-21
Update test_unit_dimension in emmocheck to EMMO 1.0.0-beta5

jesper-friis:
# Description
Update test_unit_dimension in emmocheck to EMMO 1.0.0-beta5.

A test for this PR is added in PR #679.

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [ ] Test update.

## Checklist

This checklist can be used as a help for the reviewer.

- [ ] Is the code easy to read and understand?
- [ ] Are comments for humans to read, not computers to disregard?
- [ ] Does a new feature has an accompanying new test (in the CI or unit testing schemes)?
- [ ] Has the documentation been updated as necessary?
- [ ] Does this close the issue?
- [ ] Is the change limited to the issue?
- [ ] Are errors handled for all outcomes?
- [ ] Does the new feature provide new restrictions on dependencies, and if so is this documented?

PR #679, opened 2023-11-21
Added test for get_indirect_is_a() and update to EMMO 1.0.0-beta5

jesper-friis:
# Description
Added test for get_indirect_is_a() in PR #678. 

This test is kept in a separate PR since it expects EMMO 1.0.0-beta5. The update to 1.0.0-beta5 implies some additional updates...

## Type of change

- [ ] Bug fix.
- [x] New feature.
- [ ] Documentation update.
- [ ] Test update.

## Checklist


ontopy/patch.py
```diff
@@ -99,11 +99,11 @@ def _setitem(self, name, value):
...
     >>> emmo = get_emmo()
     >>> emmo.Atom['altLabel']
-    [locstr('ChemicalElement', 'en')]
+    ['ChemicalElement']
     >>> emmo.Atom['altLabel'] = 'Element'
     >>> emmo.Atom['altLabel'] = locstr('Atomo', 'it')
     >>> emmo.Atom['altLabel']
-    [locstr('ChemicalElement', 'en'), 'Element', locstr('Atomo', 'it')]
+    ['ChemicalElement', 'Element', 'Atomo']
     """
```
francescalb, line 107: Which version of owlready2 do you have? With recent versions of owlready2 the locstr is returned.

ontopy/ontology.py
```diff
@@ -108,16 +108,16 @@ def get_ontology(
 
         if base_iri == "emmo":
             base_iri = (
-                "http://emmo-repo.github.io/versions/1.0.0-beta4/emmo.ttl"
+                "http://emmo-repo.github.io/versions/1.0.0-beta5/emmo.ttl"
```
francescalb, line 111: We should not do this, but keep the latest stable version as standard.

ontopy/ontology.py
```diff
@@ -108,16 +108,16 @@ def get_ontology(
...
             )
         elif base_iri == "emmo-inferred":
             base_iri = (
-                "https://emmo-repo.github.io/versions/1.0.0-beta4/"
+                "https://emmo-repo.github.io/versions/1.0.0-beta5/"
```
francescalb, line 115: We should not do this, but keep the latest stable version as standard.

ontopy/ontology.py
```diff
@@ -108,16 +108,16 @@ def get_ontology(
...
                 "emmo-inferred.ttl"
             )
         elif base_iri == "emmo-development":
             base_iri = (
-                "https://emmo-repo.github.io/versions/1.0.0-beta4/"
+                "https://emmo-repo.github.io/versions/1.0.0-beta5/"
```
francescalb, line 120: This is good, then we use emmo-development in the test

tests/ontopy_tests/test_patch.py
```diff
@@ -87,6 +92,13 @@ def test_get_by_label_onto(emmo: "Ontology") -> None:
     )  # Check that wikipediaReference can be acceses as attribute
 
 
+def test_get_indirect_is_a(emmo: "Ontology") -> None:
+    assert any(
+        re.match("^emmo.*.hasDimensionString.value(.*)$", str(e))
+        for e in emmo.MicroPascal.get_indirect_is_a()
+    )
+
+
```
francescalb, line 101: Why is this test put in test_get_by_label?

tests/ontopy_tests/test_patch.py
```diff
@@ -87,6 +92,13 @@ def test_get_by_label_onto(emmo: "Ontology") -> None:
     )  # Check that wikipediaReference can be acceses as attribute
 
 
+def test_get_indirect_is_a(emmo: "Ontology") -> None:
+    assert any(
```
francescalb, line 96:
```suggestion
def test_get_indirect_is_a() -> None:
    from ontopy import get_ontology
    emmo = get_ontology('emmo-development').load()
    assert any(
```

francescalb (changes requested):
I suggest updating only the default 'emmo-develeopment' to beta5 and keep the stable noes to beta4. Then we can use the development version for this updated test only.

For the other tests and code that you have changed I have created issue https://github.com/emmo-repo/EMMOntoPy/issues/680

where I have copied in your changes with updates to beta5 so that these are not lost once the upgrade to new stable version of EMMO is done.

ontopy/patch.py
jesper-friis: I have Owlready2 v0.41.

tests/ontopy_tests/test_patch.py
jesper-friis: Its not, it is put in test_patch.py

ontopy/ontology.py
jesper-friis: Fine for me. Then we can just drop this PR and skip testing get_indirect_is_a()

jesper-friis:
> I suggest updating only the default 'emmo-develeopment' to beta5 and keep the stable noes to beta4. Then we can use the development version for this updated test only.
> 
> For the other tests and code that you have changed I have created issue #680
> 
> where I have copied in your changes with updates to beta5 so that these are not lost once the upgrade to new stable version of EMMO is done.

Fine with me. The valuable part is in PR #678. This PR just tests it. Feel free to close it.

PR #682, opened 2023-11-24
Update to owlready0.45

francescalb:
# Description
1) Owlready 0.45 introduced explicit type on all literals. This needs and update in get_version.
2) python_names are added into the ontology for properties. this is not something we desire in EMMO, and hence these should be removed as default when saving in EMMOntoPy.
3) Point 2 requires a copy funtion so that the ontology itself is not changed upon saving.

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [ ] Test update.

## Checklist


francescalb:
closed with #795 and previous PRs

PR #700, opened 2024-01-16
Allow controling ontology IRI when saving

jesper-friis:
# Description
Allow separating the IRI of the ontology from `base_iri` when saving using rdflib.

Unfortunately owlready2 doesn't allow to separate the IRI of the ontology from the `base_iri` (the namespace of the entities in the ontology). In many ontologies, e.g. EMMO, are these not the same. 

Further work: Also assign the ontology IRI when loading an ontology. 

An even better solution would be to ask Jean-Baptiste to add support for separate ontology `iri` and `base_iri` to Owlready2.

## Type of change

- [ ] Bug fix.
- [x] New feature.
- [ ] Documentation update.
- [ ] Test update.

## Checklist


tests/ontopy_tests/test_iri.py
```diff
@@ -0,0 +1,23 @@
...
+# if True:
+def test_iri():
+    thisdir = Path(__file__).resolve().parent
+    ontodir = thisdir.parent / "testonto"
+    outdir = thisdir.parent / "output"
+
+    onto = get_ontology(ontodir / "testonto.ttl").load()
+    onto.base_iri = "http://example.com/onto"
+    onto.iri = "http://example.com/onto/testonto"
+    onto.save(outdir / "testonto.ttl")
```
francescalb, line 16:
```suggestion
    if os.path.exists(outdir / "testonto.ttl"):
        os.remove(outdir / "testonto.ttl")
    onto.save(outdir / "testonto.ttl")
```

tests/ontopy_tests/test_iri.py
```diff
@@ -0,0 +1,23 @@
+import re
+from pathlib import Path
```
francescalb, line 2:
```suggestion
from pathlib import Path
import os
```

francescalb (approved):
Good. I have only one suggestion, which is to remove the file you create if it is there already.
See the comments.

PR #710, opened 2024-01-26
Update save recursive and layout

francescalb:
# Description
    Update layout and fix recursive saving of ontology

    * save now returns the path to the saved ontology.
      This change does not change the previous usage.
      This is practical when the filename is generated and not given.

    * generated layout is updated so that the root is (protocol+domain)
      if there are more of these. Otherwise the generated layout is as before.

    * updated test_save to check emmo-structure, recursive saving, also for
      emmo based domain ontologies that import ontologies with various starts of iris

## Type of change

- [x] Bug fix.
- [x] New feature.
- [x] Documentation update.
- [x] Test update.

## Checklist


ontopy/ontology.py
```diff
@@ -891,17 +891,26 @@ def save(
...
         write_catalog_file: bool
             Whether to also write a catalog file to disk.
         append_catalog: bool
             Whether to append to an existing catalog file.
         catalog_file: str | Path
             Name of catalog file.  If not an absolute path, it is prepended
             to `dir`.
+
+        Returns:
+        ----------
```
jesper-friis, line 912:
```suggestion
        Returns
        -------
```

ontopy/ontology.py
```diff
@@ -922,36 +931,44 @@ def save(
...
-                filename = f"{self.name}.{fmt}"
+                file = f"{self.name}.{fmt}"
             else:
                 raise TypeError("`filename` and `format` cannot both be None.")
-        filename = os.path.join(dir, filename)
-        dir = Path(filename).resolve().parent
+        else:
+            file = filename
+        filepath = os.path.join(dir, file)
+        returnpath = filepath
```
jesper-friis, line 944: One could consider making `filepath` a pathlib.Path object. That would also mean that save() will return a Path object instead of a string, which would be little more explicit and potentially little more cross-platform.

ontopy/ontology.py
```diff
@@ -984,13 +1001,12 @@ def save(
...
             write_catalog(
-                {self.get_version(as_iri=True): filename},
+                {self.get_version(as_iri=True): filepath},
                 output=catalog_file,
                 directory=dir,
                 append=append_catalog,
             )
-
-        if squash:
+        elif squash:
```
jesper-friis, line 1009: This means that if you specify `write_catalog_file` you will not be able to squash. I agree that it is not a very sensible to write a catalogue file when squashing, but why prohibit it?

```suggestion
        if squash:
```

ontopy/ontology.py
```diff
@@ -1010,9 +1026,9 @@ def save(
                     graph.remove((s, p, o))
                     graph.add((URIRef(self.iri), p, o))
 
-            graph.serialize(destination=filename, format=format)
+            graph.serialize(destination=filepath, format=format)
```
jesper-friis, line 1029: Seems that we should add a test that saves a squashed ontology...

ontopy/utils.py
```diff
@@ -786,6 +786,14 @@ def directory_layout(onto):
 
         where `ontoA`, `ontoB` and `ontoC` are imported Ontology objects.
     """
+    all_imported = [
+        imported.base_iri for imported in onto.indirectly_imported_ontologies()
+    ]
+    # get protocol and domain of all imported ontologies
+    namespace_roots = set()
+    for iri in all_imported:
+        protocol, domain, *_ = urllib.parse.urlsplit(iri)
```
jesper-friis, line 795: Is the syntax `*_` valid in all versions of Python that we want to support?

tests/test_save.py
```diff
@@ -65,6 +65,9 @@ def test_save(
     # 4. save testonto to testonto.owl again, but with overwrite=True
     # 5. check that testonto.owl is the same as testonto_saved.owl
     # NB! this is not currently working, issue #685
+    # It might be that this inetnional behaviour of save should be changed.
```
jesper-friis, line 68:
```suggestion
    # It might be that this intentional behaviour of save should be changed.
```

tests/test_ontology_squash.py
jesper-friis, line 1: Why is this test removed?

jesper-friis (commented):
Very nice PR :+1: 

Added a few comments and suggestions.

ontopy/ontology.py
francescalb: The problem is then that when it does a recursive save it will then proceed through this next if/else part and actually write self again (in the wrong place). You can try.

ontopy/ontology.py
francescalb: You made that at the beginning of this week....

ontopy/utils.py
francescalb: That is a typo. I'll remove the star.

francescalb: Nope, it was not a typo. Apparently this should work...

tests/test_ontology_squash.py
francescalb: It was moved to be in the other files where ontology_save is tested

ontopy/ontology.py
francescalb: Done

ontopy/ontology.py
jesper-friis: But we explicitly set `squash=False` when `save()` calls itself recursively.

ontopy/ontology.py
jesper-friis: Codecov just marked that squashing was not tested. But that is gone now, so the test seems to have been readded now.

ontopy/utils.py
jesper-friis: I remember seeing this syntax being described in a release note for a new Python release. But I can't remember what release. What is the oldest version of Python we support?

jesper-friis (commented):
Followed up on two comments

ontopy/utils.py
francescalb: It passes in all the tests....

jesper-friis (approved):

PR #715, opened 2024-02-12
Updated ontoconvert help

jesper-friis:
# Description
Seems that the `--base-iri` option of `ontoconvert` is not implemented. Added a note about that in the `--help` documentation.

This option is used in some EMMO workflows, so we keep it to not break them.

This is a place where we really should have added a "TODO" comment in the code - but pylint doesn't accept that.

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [x] Documentation update.
- [ ] Test update.

## Checklist


tools/ontoconvert
```diff
@@ -86,13 +86,15 @@ def main(argv: list = None):
         action="store_true",
         help="Do not infer imported ontologies.",
     )
+    # To be implemented...
```
francescalb, line 89:
```suggestion
    # To be implemented...Write an issue about it and refer to that issue here
```

jesper-friis: Added issue https://github.com/emmo-repo/EMMOntoPy/issues/716

PR #719, opened 2024-02-25
Correct saving squashed ontology

jesper-friis:
# Description
Added `--iri` and `--base-iri` options to ontoconvert allowing to correctly saving a squashed ontology.

This is both a fix in EMMOntoPy and a workaround for the bug in Owlready2 that replaces the final hash (#) of the `base_iri` to a slash (/).

Test it with:

    cd EMMO
    git fetch origin 1.0.0-beta7:1.0.0-beta7
    git co 1.0.0-beta7
    ontoconvert -wsa -b "https://w3id.org/emmo#" -I "https://w3id.org/emmo" emmo.ttl emmo-squashed.ttl

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [ ] Test update.

## Checklist


francescalb (changes requested):
Looks good as far as I was able to see when testing. However, tests are missing. Maybe add a test where testonto.ttl is squash saved and then check that this is the same as the original after loading? First check that new IRI and BASE_IRI are changed and change back of course...
And also check that the squashed ontology no longer imports other ontologies (Is this correct?)

jesper-friis:
> Looks good as far as I was able to see when testing. However, tests are missing. Maybe add a test where testonto.ttl is squash saved and then check that this is the same as the original after loading? First check that new IRI and BASE_IRI are changed and change back of course... And also check that the squashed ontology no longer imports other ontologies (Is this correct?)
> 
> Also, where is / --> # hack done? Is it a good idea to move this fix out of squash?

Added a second test with squash and change of IRIs to tests/tools/test_ontoconvert.py

There is no explicit / --> # hack. Instead the --iri and --base-iri options of the ontoconvert tool allow setting the iri and base_iri after the ontology is loaded. Hence, these options can be used to work around the bug in Owlready2.

francescalb:
> > Looks good as far as I was able to see when testing. However, tests are missing. Maybe add a test where testonto.ttl is squash saved and then check that this is the same as the original after loading? First check that new IRI and BASE_IRI are changed and change back of course... And also check that the squashed ontology no longer imports other ontologies (Is this correct?)
> > Also, where is / --> # hack done? Is it a good idea to move this fix out of squash?
> 
> Added a second test with squash and change of IRIs to tests/tools/test_ontoconvert.py
> 
> There is no explicit / --> # hack. Instead the --iri and --base-iri options of the ontoconvert tool allow setting the iri and base_iri after the ontology is loaded. Hence, these options can be used to work around the bug in Owlready2.

So the bug is still there for any other way of saving the ontology (i.e. not using ontoconvert)?

francescalb (changes requested):
Thank you for updating the test. Since you have decided not to use pylint eand fixtures a clean up of that is neded.

jesper-friis:
> > > Looks good as far as I was able to see when testing. However, tests are missing. Maybe add a test where testonto.ttl is squash saved and then check that this is the same as the original after loading? First check that new IRI and BASE_IRI are changed and change back of course... And also check that the squashed ontology no longer imports other ontologies (Is this correct?)
> > > Also, where is / --> # hack done? Is it a good idea to move this fix out of squash?
> > 
> > 
> > Added a second test with squash and change of IRIs to tests/tools/test_ontoconvert.py
> > There is no explicit / --> # hack. Instead the --iri and --base-iri options of the ontoconvert tool allow setting the iri and base_iri after the ontology is loaded. Hence, these options can be used to work around the bug in Owlready2.
> 
> So the bug is still there for any other way of saving the ontology (i.e. not using ontoconvert)?

Yes, it is still there. Issue #723 is a suggestion to work around it when loading an ontology.

ontopy/testutils.py
```diff
@@ -0,0 +1,29 @@
+"""Motule primarly intended to be imported by tests.
```
francescalb, line 1:
```suggestion
"""Module primarly intended to be imported by tests.
```

francescalb (approved):
Approved, just a minor typo to be fixed.

PR #732, opened 2024-03-14
Add --copy-annotation option to ontoconvert

jesper-friis:
# Description
This option is needed to automatically rdfs:label annotations to classes and properties when publishing EMMO to increase the FAIRness or the ontology.

## Type of change
- [ ] Bug fix.
- [x] New feature.
- [ ] Documentation update.
- [ ] Test update.

## Checklist

### DIFF
--- a/ontopy/utils.py
+++ b/ontopy/utils.py
@@ -840,7 +840,11 @@ def directory_layout(onto):
 def copy_annotation(onto, src, dst):
     """In all classes and properties in `onto`, copy annotation `src` to `dst`.
 
-    The `src` and `dst` can either be provided as a label string or a full IRI.
+    Arguments:
+        onto: Ontology to work on.
+        src: Name of source annotation.
+        dst: Name or IRI of destination annotation.  Use IRI if the
+            destination annotation is not already in the ontology.
     """
     if onto.world[src]:
         src = onto.world[src]
--- a/ontopy/utils.py
+++ b/ontopy/utils.py
@@ -861,7 +861,7 @@ def copy_annotation(onto, src, dst):
                 "new destination annotation property must be provided as "
                 "a full IRI"
             )
-        name = min(dst.rsplit("#")[-1], dst.rsplit("/")[-1])
+        name = min(dst.rsplit("#")[-1], dst.rsplit("/")[-1], key=len)
         iri = dst
         dst = onto.new_annotation_property(name, owlready2.AnnotationProperty)
         dst.iri = iri
--- a/ontopy/utils.py
+++ b/ontopy/utils.py
@@ -835,3 +835,34 @@ def directory_layout(onto):
                     )
 
     return layout
+
+
+def copy_annotation(onto, src, dst):
+    """In all classes and properties in `onto`, copy annotation `src` to `dst`.
+
+    The `src` and `dst` can either be provided as a label string or a full IRI.
+    """
+    if onto.world[src]:
+        src = onto.world[src]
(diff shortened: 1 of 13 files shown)


ontopy/utils.py
```diff
@@ -835,3 +835,34 @@ def recur(o):
...
     return layout
+
+
+def copy_annotation(onto, src, dst):
+    """In all classes and properties in `onto`, copy annotation `src` to `dst`.
+
+    The `src` and `dst` can either be provided as a label string or a full IRI.
+    """
+    if onto.world[src]:
+        src = onto.world[src]
```
francescalb, line 846: Why check in world?

ontopy/utils.py
```diff
@@ -835,3 +835,34 @@ def recur(o):
...
+    else:
+        src = onto[src]
+
+    if onto.world[dst]:
+        dst = onto.world[dst]
+    elif dst in onto:
+        dst = onto[dst]
+    else:
+        if "://" not in dst:
+            raise ValueError(
```
francescalb, line 856: I think we should have tests for errors that we expect may happen.

ontopy/utils.py
```diff
@@ -835,3 +835,34 @@ def recur(o):
...
+                "new destination annotation property must be provided as "
+                "a full IRI"
+            )
+        name = min(dst.rsplit("#")[-1], dst.rsplit("/")[-1])
+        iri = dst
+        dst = onto.new_annotation_property(name, owlready2.AnnotationProperty)
+        dst.iri = iri
```
francescalb, line 863: Needs test

tests/tools/test_ontoconvert.py
```diff
@@ -39,3 +39,18 @@ def test_run() -> None:
     assert re.search("@prefix : <https://w3id.org/ex/testonto#>", output2)
     assert re.search("<https://w3id.org/ex/testonto> .* owl:Ontology", output2)
     assert re.search("testclass .* owl:Class", output2)
+
+    # Test 3 - copy-annotation
```
francescalb, line 43: you are checking copy_preflabel. I agree that this is an alias for annotation, but I think it would be good to check that other annotations also are added. Most particularly, that we can copy to an  annotation that is originally not in the ontology.

Also, I think we should have a test where preflabel is not converted, as prefLabel is copied by default (if I understand store_true correctly).

ontopy/utils.py
```diff
@@ -835,3 +835,34 @@ def recur(o):
                     )
 
     return layout
+
+
+def copy_annotation(onto, src, dst):
+    """In all classes and properties in `onto`, copy annotation `src` to `dst`.
+
+    The `src` and `dst` can either be provided as a label string or a full IRI.
```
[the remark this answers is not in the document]
jesper-friis: done

ontopy/utils.py
jesper-friis: Because we want to support src of the form `http://www.w3.org/2004/02/skos/core#prefLabel`

ontopy/utils.py
jesper-friis: Isn't this such a test? 

What other errors would you expect?

ontopy/utils.py
jesper-friis: This is already tested in tests/tools/test_ontoconvert.py

Just codecov that doesn't understand tests done in subprocesses...

francescalb (approved):
Nice

PR #740, opened 2024-04-03
Make it possible to run HermiT on EMMO

jesper-friis:
# Description
Make it possible to run HermiT on EMMO, by:
* Add support for additional keyword arguments that `sync_reasoner()` may pass to the `save()` method.
* Remove custom datatypes before running HermiT. Readd the removed data properties after HermiT completes.
* Add `--java-executable` and `--java-memory` arguments to ontoconvert to control how to run java.

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [ ] Test update.

## Checklist


ontopy/ontology.py
```diff
@@ -940,7 +942,9 @@ def save(
                 raise TypeError("`filename` and `format` cannot both be None.")
         else:
             file = filename
-        filepath = os.path.join(dir, file)
+        filepath = os.path.join(
+            dir, file if isinstance(file, str) else file.name
+        )
```
francescalb, line 947: why?

francescalb:
We should have a test on this. It is important that we wuickly pick up if running sync_reasoner on EMMO does not work.

ontopy/ontology.py
jesper-friis: Add fix: `isinstance(file, str)` -> `isinstance(file, (str, Path))`

francescalb (approved):
As a general rule I think that untested features should be tested properly before being smerged to master.

Please make an issue that clearly decribes what needs to be tested so that it can be picked up later.

PR #741, opened 2024-04-04
Fix failing test

jesper-friis:
# Description
Replaced loading battinfo with testonto to avoid failures from battinfo that currently doesn't load properly.

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [ ] Test update.

## Checklist

### DIFF
--- a/tests/test_get_by_label.py
+++ b/tests/test_get_by_label.py
@@ -103,20 +103,22 @@ def test_get_by_label_emmo(emmo: "Ontology") -> None:
     from emmopy import get_emmo
     from ontopy import get_ontology
     from ontopy.ontology import NoSuchLabelError
+    from ontopy.testutils import ontodir
 
     emmo = get_emmo()
     assert emmo[emmo.Atom.name] == emmo.Atom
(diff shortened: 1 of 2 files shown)


francescalb (changes requested):
No, this is to circumvent the problem. 

The problem is that the catalog is not updated for each import, which means that loading battery.ttl directly works nicely because all the imported ontologies are included in the "parent" catalog-file. This is however, not the case for battinfo, and thus it cannot fine the subontologies.

francescalb:
> No, this is to circumvent the problem.
> 
> The problem is that the catalog is not updated for each import, which means that loading battery.ttl directly works nicely because all the imported ontologies are included in the "parent" catalog-file. This is however, not the case for battinfo, and thus it cannot fine the subontologies.

No, I correct myself. The problem is that the original catalogfile of battinfo is not updated correctly (battery.ttl in master is now at version 0.9). This is dangerous...

francescalb:
Clsoed with #745

PR #742, opened 2024-04-05
Rewriting ontodoc based on domain-battery

jesper-friis:
# Description
Rewriting and simplifying the ontodoc tool based on domain-battery. For now the python module is called ontodoc_rst.py.

Structure the generated document as follows:

```
==========
References
==========

Module: <first_module_name>
===========================

-------
Classes
-------

<class1>
--------

...
```

For each module, document:
- classes
- object properties
- data properties
- annotation properties
- individuals
- datatypes

## Tasks
- [x] Reproduce all information included in the [battery.rst](https://github.com/emmo-repo/domain-battery/blob/master/sphinx/battery.rst) file
- [x] Add tooling for creating other needed files, like `index.rst` (just a template that the user can improve)
- [x] Add a new ontodoc script/tool - instead of adding a new tool, the old ontodoc tool was extended to support `.rst` output. It doesn't run sphinx automatically, though...
- [ ] Add small tutorial of how to use the tool
- [ ] Add ontology metadata documentation to generated documentation
- [ ] Add more semantic information to generated documentation
- [ ] Add language to keys for localised values
- [ ] Add default css file (common for all EMMO domain ontologies, could e.g. be downloaded from a common source...)
- [ ] Add support for figures (e.g. by displaying links to known image formats as images)
- [ ] Sort modules in a logical order, e.g. by namespace and import depth
- [ ] Add a module dependency graph at the top
- [ ] Add subclasses to generated documentation to support navigating the taxonomy
- [ ] Is it possible to collapse the sidebar to make it easier to navigate to the right module?
- [ ] Add anchor by prefLabel, to make it easier for people to find the IRI when one knows the prefLabel

The last points can maybe be done in a separate PR.

## Type of change
- [ ] Bug fix.
- [x] New feature.
- [ ] Documentation update.
- [x] Test update.

## Checklist


tests/test_ontodoc_rst.py
```diff
@@ -0,0 +1,20 @@
...
+# if True:
+def test_ontodoc():
+    """Test ontodoc."""
+    from pathlib import Path
+
+    from ontopy import get_ontology
+    from ontopy.ontodoc_rst import OntologyDocumentation
+    import owlready2
+
+    emmo = get_ontology("https://w3id.org/emmo/1.0.0-rc1").load()
```
francescalb, line 13: I am not so happy about tests that dowload EMMO. We already have many of those and they make the tests very slow.

tests/tools/test_ontodoc.py
```diff
@@ -48,3 +48,21 @@ def test_run_w_punning() -> None:
...
+
+    ontodoc.main(
+        [
+            "--imported",
+            "--reasoner=HermiT",
+            "--iri-regex=^https://w3id.org/emmo/domain",
+            str(test_file),
+            str(outdir / "mammal.rst"),
+        ]
+    )
```
francescalb, line 68: Does this test what is actually created?

tests/test_save.py
```diff
@@ -238,16 +224,16 @@ def test_save_emmo_domain_ontology(
     assert set(
         os.listdir(outputdir / "emmo.info" / "emmo" / "domain" / "chameo")
     ) == {"chameo.rdfxml", "catalog-v001.xml"}
-    assert set(
+    assert not {"isq.rdfxml", "catalog-v001.xml"}.difference(
         os.listdir(outputdir / "emmo.info" / "emmo" / "disciplines")
-    ) == {"isq.rdfxml", "catalog-v001.xml"}
```
francescalb, line 243: Why?

tests/test_save.py
```diff
@@ -263,8 +249,8 @@ def test_save_emmo_domain_ontology(
...
     assert set(
-        os.listdir(outputdir / "emmo.info" / "emmo" / "domain" / "chameo")
+        os.listdir(outputdir2 / "emmo.info" / "emmo" / "domain" / "chameo")
     ) == {"chameo.rdfxml", "catalog-v001.xml"}
-    assert set(
-        os.listdir(outputdir / "emmo.info" / "emmo" / "disciplines")
-    ) == {"isq.rdfxml", "catalog-v001.xml"}
+    assert not {"isq.rdfxml", "catalog-v001.xml"}.difference(
+        os.listdir(outputdir2 / "emmo.info" / "emmo" / "disciplines")
+    )
```
francescalb, line 256: Why this? It is more difficult to read.

ontopy/utils.py
```diff
@@ -94,12 +97,45 @@ def isinteractive():
...
+    >>> ontopy.utils.PREFERRED_LANGUAGE = "en"
+
+    """
+    for langstr in langstrings:
+        if hasattr(langstr, "lang") and langstr.lang == PREFERRED_LANGUAGE:
+            return langstr
+    for langstr in langstrings:
+        if not hasattr(langstr, "lang"):
+            return langstr
+    return langstr[0]
```
francescalb, line 119: This is tested very indirectly. Isnæt it better to test this in a unit test?

tests/test_save.py
jesper-friis: Because onto.save() stores other files in addition to isq.rdfxml and the catalogue file. Explicitly checking for all created files will fail when new modules are added to EMMO disciplines.

tests/test_save.py
jesper-friis: I agree that it is more difficult to read. But without this the test is failing since onto.save() creates other files in addition to isq.rdfxml and the catalogue file. 

We can either explicitly checking for all created files, but that will fail when new modules are added to EMMO disciplines.

Alternatively we could rewrite the test as

```python
created_files = set(os.listdir(outputdir2 / "emmo.info" / "emmo" / "disciplines"))
for fname in ("chameo.rdfxml", "catalog-v001.xml"):
    assert fname in created_files
```

tests/tools/test_ontodoc.py
jesper-friis: Added simple tests for generated content...

tests/test_ontodoc_rst.py
jesper-friis: Replaced it with the local animal ontology.

ontopy/utils.py
jesper-friis: Added explicit test

tests/test_save.py
francescalb: yes, I prefer this.

francescalb (approved):
Approved, but I prefer the change in one of the tests as you propose in the comment

tests/test_save.py
jesper-friis: done

PR #756, opened 2024-05-01
Fixed failing test_save in master

jesper-friis:
# Description
Simplifying paths in test_save_emmo_domain_ontology() in test_save.py to make it much easier to debug.


francescalb (changes requested):
There is no bug in the current master. I passes without problem.

jesper-friis:
> There is no bug in the current master. I passes without problem.

This PR is just a simplification without any behavioral changes, so it shouldn't change whether the tests fails or not. Still it made `test_save.py` succeed locally for me. Merging it into PR #742, also helped that PR succeeding on github.

francescalb (approved):
OK, since you insist. But please rewrite this to a change because of your preference as there is no bug here.

PR #767, opened 2024-06-02
Added figures to generated documentation

jesper-friis:
Added figures to generated documentation. If the value of a property matches a regex for a figure on the web, it will be included in the documentation. 

**Further work**: The figure size should probably be adjusted to not make the table too large. Probably preferred to only show a thumbnail-sized version of the figure in the table. The full size could be shown by clicking. Easier to play with in a following-up PR when this is applied to EMMO.

---
Also, added prefLabel as anchoring points such that one can use the browser to easily get documentation of a concept one knows the prefLabel of.

For example, entering

    https://w3id.org/emmo#Atom

should bring you to the section in the html doc that documents Atom.

# Description

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [ ] Test update.

## Checklist


francescalb (changes requested):
I see no example or test. 

At least I want t more through description in the code itself. E.g. adapt text in the PR comment and include it in the documentation.

jesper-friis:
> I see no example or test.
> 
> At least I want t more through description in the code itself. E.g. adapt text in the PR comment and include it in the documentation.

It is tested - both online in production and in the tests/ folder

jesper-friis:
> I see no example or test.
> 
> At least I want t more through description in the code itself. E.g. adapt text in the PR comment and include it in the documentation.

It is tested, both online and in the tests/ folder

PR #795, opened 2024-10-29
Removed owlready2:python_names in genated triples in excelparser

francescalb:
# Description

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [ ] Test update.

## Checklist

### DIFF
--- a/ontopy/excelparser.py
+++ b/ontopy/excelparser.py
@@ -456,6 +456,15 @@ def create_ontology_from_pandas(  # pylint:disable=too-many-locals,too-many-bran
     entities_with_errors = {
         key: set(value) for key, value in entities_with_errors.items()
     }
+
+    # Remove triples with predicate 'python_name' added by owlready2i>0.45
+    onto._del_data_triple_spod(  # pylint: disable=protected-access
+        p=onto._abbreviate(  # pylint: disable=protected-access
+            "http://www.lesfleursdunormal.fr/static/_downloads/"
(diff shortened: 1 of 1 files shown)


ontopy/excelparser.py
```diff
@@ -456,6 +456,15 @@ def create_ontology_from_pandas(  # pylint:disable=too-many-locals,too-many-bran
...
         key: set(value) for key, value in entities_with_errors.items()
     }
+
+    # Remove triples with predicate 'python_name' added by owlready2i>0.45
+    onto._del_data_triple_spod(  # pylint: disable=protected-access
+        p=onto._abbreviate(  # pylint: disable=protected-access
+            "http://www.lesfleursdunormal.fr/static/_downloads/"
+            "owlready_ontology.owl#python_name"
+        )
+    )
```
jesper-friis, line 466: If `create_ontology_from_pandas()` is called with an `input_ontology`, then this will modify the input ontology. I suggest that you on line 314 replaces `onto = input_ontology` with onto = input_ontology.copy()` to ensure that we are not modifying the input ontology.

jesper-friis (commented):
See the comment. Otherwise, looks fine.

ontopy/excelparser.py
francescalb: Hmmm, in that case we should not remove the triples with python_name coming from the input_ontology...

jesper-friis (approved):

PR #809, opened 2024-11-24
Updated emmocheck to ahead to latest formulation of units

jesper-friis:
# Description
Updated emmocheck to ahead to latest formulation of units

**Note**: this PR is already in use when testing the development version of EMMO. When merging this PR, please update the pip install EMMOntoPy statement in https://github.com/emmo-repo/EMMO/blob/master/.github/workflows/ci_emmocheck.yml

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [ ] Test update.

## Checklist


emmopy/emmocheck.py
```diff
@@ -500,7 +508,7 @@ def test_quantity_dimension(self):
                             issubclass(cls, self.onto.ISQDimensionlessQuantity)
                         )
 
-    def test_dimensional_unit(self):
+    def test_dimensional_unit_rc2(self):
```
francescalb, line 511: Why do we have two tests? Shuold we not just stick to one?

jesper-friis: If I only had updated the test to the latest version of EMMO without keeping a copy for the earlier version, you would probably not have commented on it. This is how most development handle changes in their dependencies. If you use an old version of EMMO, you simply test it with an older version of EMMOntoPy.

To make the latest version of EMMOntoPy handle multiple versions of EMMO, it is not sufficient to rename the tests differently, we need a whole new formalisation to (preferably in a declarative way) specify what versions of EMMO each test is valid for. I added issue #821 for that.

francescalb (approved):

PR #810, opened 2024-11-24
Updated sync_reasoner() such that it works for FaCT++

jesper-friis:
# Description
Updated sync_reasoner() such that it works for FaCT++.

This update makes it possible to reason on EMMO with FaCT++. Unfortunately EMMO 1.0.0-rc3 refer to unsupported datatypes without declaring them as of type rdfs:Datatype. This means that we have to hardcode the unsupported datatypes. 

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [ ] Test update.

## Checklist

### DIFF
--- a/ontopy/ontology.py
+++ b/ontopy/ontology.py
@@ -1343,67 +1343,85 @@ class Ontology(owlready2.Ontology):  # pylint: disable=too-many-public-methods
 
         Keyword arguments are passed to the underlying owlready2 function.
         """
-        # pylint: disable=too-many-branches
-
-        removed_equivalent = defaultdict(list)
-        removed_subclasses = defaultdict(list)
+        # pylint: disable=too-many-branches,too-many-locals
+        # pylint: disable=unexpected-keyword-arg
+        removed_gspo = []  # obj: (ontology, s, p, o)
+        removed_gspod = []  # data: (ontology, s, p, o, d)
 
         if reasoner == "FaCT++":
             sync = sync_reasoner_factpp
+            remove_custom_datatypes = True
         elif reasoner == "Pellet":
             sync = owlready2.sync_reasoner_pellet
+            remove_custom_datatypes = False
         elif reasoner == "HermiT":
             sync = owlready2.sync_reasoner_hermit
-
-            # Remove custom data propertyes, otherwise HermiT will crash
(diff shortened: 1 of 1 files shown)


ontopy/ontology.py
```diff
@@ -1343,67 +1343,85 @@ def sync_reasoner(
...
-                        if (
-                            hasattr(r.value.__class__, "iri")
-                            and r.value.__class__.iri == datatype_iri
-                        ):
-                            remove_subcls.append(i)
-                            removed_subclasses[cls].append(r)
-                for i in reversed(remove_subcls):
-                    del cls.is_a[i]
-
+            remove_custom_datatypes = False
```
francescalb, line 1359: Doesn't HermiT crash if datatypes are not custom datatypes any more?

francescalb (commented):
I am a bit unconfortable with a PR that changes something that apparently is an issue but that we did not catch in our tests. What is your opinion that in this specific case?

jesper-friis:
> I am a bit unconfortable with a PR that changes something that apparently is an issue but that we did not catch in our tests. What is your opinion that in this specific case?

I think it would be too time-consuming to run FaCT++ and HermiT on EMMO in the GitHub tests. For now I think it is fine with no CI tests, since this is a feature that we need and will test anyway in the publication of EMMO.

When we get time, we could consider to adding datatypes to a test ontology and run the reasoners on that.

francescalb:
> > I am a bit unconfortable with a PR that changes something that apparently is an issue but that we did not catch in our tests. What is your opinion that in this specific case?
> 
> I think it would be too time-consuming to run FaCT++ and HermiT on EMMO in the GitHub tests. For now I think it is fine with no CI tests, since this is a feature that we need and will test anyway in the publication of EMMO.
> 
> When we get time, we could consider to adding datatypes to a test ontology and run the reasoners on that.

We already have tests that run on github. But apparently not good enough.

francescalb (approved):
Approved, provided a godd issue of what is needed to add in upated tests for reasoning is provided.

jesper-friis:
Updated sync_reasoner() such that it works for FaCT++.

This update makes it possible to reason on EMMO with FaCT++. Unfortunately EMMO 1.0.0-rc3 refer to unsupported datatypes without declaring them as of type rdfs:Datatype. This means that we have to hardcode the unsupported datatypes.

PR #814, opened 2024-12-04
Importing rdfs schemas

### LINKED ISSUE #797: The test for load_foaf is bypassed.
Check out.

### LINKED ISSUE #811: Loading non-emmo-related ontologies
We wanted to load a schema.org schema (https://schema.org/version/latest/schemaorg-current-https.ttl) using `get_ontology().load()`, but it returns an empty object. If we change the "Class" prefix to "owl" and the "rdfs:label" to "skos:prefLabel" it does load all the classes. The only difference with loading an emmo ontology is that the label has to be retrieved from `get_preferred_label()` instead of "prefLabel". Would be convenient if other ontologies could also be loaded directly from the original turtle files.

An example on what we had to change:

schema:AMRadioChannel a rdfs:Class ;
    rdfs:label "AMRadioChannel" ;
    
 to:
 
 schema:AMRadioChannel a owl:Class ;
    skos:prefLabel "AMRadioChannel" ;

francescalb:
# Description

Partially closes #811.
Closes #797 

I is not possible to easily access classes of typ rdfs:Class. This is needed if working simultaneously with an rdf-schema (as opposed to an owl ontology). Owlready2 does not search for enties that are not of type owl:XXX.
A search for rdfs:Class is now added so that when asking for classes in an ontology also the rdfs:Class'es are returned as Python objects.

This is not as readily fixed with rdfs:Property. As a first fix, such properties are now returned as a list of strings (instead of a list of python objcets). Note that the nodes are present in the onotology, but the rdfs:Poerty is not cast as a python object. 

Owlready2 is designed to support owl ontologies (not rdfs). If we really want to extend the suport for rdfs:Property the best solution is most likely to cast it to an owl:Annotation_property. If so this should be done in a next PR .


ontopy/ontology.py
```diff
@@ -1258,6 +1273,40 @@ def annotation_properties(self, imported=False):
...
+                are also returned.
+        """
+        generator = []
+        for prop in list(
+            self._entities("object_properties", imported=imported)
+        ):
+            generator.append(prop)
+
+        for prop in list(
+            self._entities("annotation_propertoes", imported=imported)
```
francescalb, line 1292:
```suggestion
            self._entities("annotation_properties", imported=imported)
```

ontopy/ontology.py
```diff
@@ -1207,9 +1211,18 @@ def _entities(
                 elif entity_type == "annotation_properties":
                     for prop in list(onto.annotation_properties()):
                         generator.append(prop)
+                elif entity_type == "properties":
```
jesper-friis, line 1214: Include "properties" as a possible value to the docstring of the `entity_type` argument.

PR #820, opened 2024-12-12
Fixed failures in rename_iris()

jesper-friis:
# Description
The `--rename-iris` is now actively used to generate `emmo-for-humans.ttl`.

**Note**: this PR is already in use for generation of https://w3id.org/emmo/emmo-for-humans, a version of EMMO that is only intended for examples which has all numerical IDs replaced with prefLabel's. 
When merging this PR, please update the `pip install EMMOntoPy` statement in https://github.com/emmo-repo/EMMO/blob/master/.github/workflows/cd_update_pages.yml

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [ ] Test update.

## Checklist


francescalb (changes requested):
Describe what this PR is about so that it is possible to do an informed review.

Also,  instead of commenting out tests, the code should be fixed so that they pass.

jesper-friis:
Closing this PR. Seems not to be needed if we call rename IRIs on a non-squashed ontology.
There might still be an issue though...

PR #839, opened 2025-03-14
Update to emmo 1.0.0 as it is now released

francescalb:
# Description

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [ ] Test update.

## Checklist


ontopy/ontology.py
```diff
@@ -564,23 +559,39 @@ def remove_label_annotation(self, iri):
...
         self,
-        iri_base: str = "http://emmo.info/emmo",
+        iri_base: str = "https://w3id.org/emmo/inferred",
         prefix: str = "emmo",
         visited: "Optional[Set]" = None,
     ) -> None:
         """Set a common prefix for all imported ontologies
         with the same first part of the base_iri.
 
+        Note that this function might give unintened results.
```
francescalb, line 569:
```suggestion
        Note that this function might give unintended results.
```

ontopy/ontology.py
```diff
@@ -564,23 +559,39 @@ def remove_label_annotation(self, iri):
...
+        I.e. if `https://w3id.org/emmo` is given as iri_base
+        all imported domain ontologies will be given the same
```
francescalb, line 571:
```suggestion
        all imported domain ontologies adhering to the emmo standard for 
        base_iri will be given the same
```

ontopy/ontology.py
```diff
@@ -564,23 +559,39 @@ def remove_label_annotation(self, iri):
...
+        prefix as well as emmo itself.
+
+        The default is to set the prefix to emmo for emmo-inferred.
```
francescalb, line 574:
```suggestion
        The default is to set the prefix to emmo only for emmo-inferred.
```

tests/test_excelparser/result_ontology/fromexcelonto.ttl
```diff
@@ -1,6 +1,6 @@
 @prefix : <http://emmo.info/emmo/domain/onto#> .
 @prefix dcterms: <http://purl.org/dc/terms/> .
-@prefix emmo: <http://emmo.info/emmo#> .
+@prefix ns1: <https://w3id.org/emmo#> .
```
jesper-friis, line 3: Wouldn't it be better to keep `emmo` as prefix?

```suggestion
@prefix emmo: <https://w3id.org/emmo#> .
```

tests/test_excelparser/result_ontology/fromexcelonto.ttl
```diff
@@ -12,19 +12,19 @@
...
     owl:versionInfo "0.01"@en .
 
 :EMMO_0264be35-e8ad-5b35-a1a3-84b37bde22d1 a owl:Class ;
-    emmo:EMMO_967080e5_2f42_4eb2_a3a9_c58143e835f9 "Temporal pattern occurring in a time interval"@en ;
-    emmo:EMMO_b432d2d5_25f4_4165_99c5_5935a7763c1a "Light house during one night"@en ;
     rdfs:subClassOf [ a owl:Restriction ;
-            owl:onProperty emmo:EMMO_e1097637_70d2_4895_973f_2396f04fa204 ;
-            owl:someValuesFrom emmo:EMMO_d4f7d378_5e3b_468a_baa1_a7e98358cda7 ],
+            owl:onProperty ns1:EMMO_e1097637_70d2_4895_973f_2396f04fa204 ;
+            owl:someValuesFrom ns1:EMMO_d4f7d378_5e3b_468a_baa1_a7e98358cda7 ],
```
jesper-friis, line 22:
```suggestion
            owl:onProperty emmo:EMMO_e1097637_70d2_4895_973f_2396f04fa204 ;
            owl:someValuesFrom emmo:EMMO_d4f7d378_5e3b_468a_baa1_a7e98358cda7 ],
```

tests/test_excelparser/result_ontology/fromexcelonto_only_classes.ttl
```diff
@@ -1,6 +1,6 @@
 @prefix : <http://emmo.info/emmo/domain/onto#> .
 @prefix dcterms: <http://purl.org/dc/terms/> .
-@prefix emmo: <http://emmo.info/emmo#> .
+@prefix ns1: <https://w3id.org/emmo#> .
```
jesper-friis, line 3: Like above. Better to keep `emmo` as prefix. 
```suggestion
@prefix emmo: <https://w3id.org/emmo#> .
```

jesper-friis (commented):
Very nice. See suggestions about choice of prefix.

tests/test_excelparser/result_ontology/fromexcelonto.ttl
francescalb: It is just the prefix in a test. And   think this requires some more thinking. With all emmo-domain ontologies having the same beginning of the base_iri it is not so straightforward to set the prefix to emmo for just emmo and not the domain ontologies. 

If we do as before, the excelparser in the test wil work as before, but that is just because we are creating an ontology with a different base_iri all toghether.

jesper-friis: No need to treat doamin ontologies separately. If the doamin ontologies uses `https://w3id.org/emmo#` as prefix for the classes and properties, they should be prefixed with `emmo:` as well.

francescalb: https://w3id.org/emmo/domain-xyz starts with the same and should not have the same prefix. 
and all emmo modules also start with the same. There is not well-described rule that differentiates between an ontology-module that is part of emmo-core and one that is not.

jesper-friis (approved):
Please create a new issue about cleaning up how we define namespaces. We need to clearly separate between the IRI of the ontology itself and the base IRI for all the entities in the ontology.

PR #845, opened 2025-03-20
Update Metadata so that all metadata-items can be accessed

### LINKED ISSUE #841: How to access metadata such as abstract and creators?
```
onto  = get_ontology('emmo.ttl').load()
onto.metadata 
```
returns and a Metadata-instance, which is missing a lot of the metadata that is actually present in the emmo-ontology, e.g. the abstract is missing.

francescalb:
# Description
closes  #841 

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [x] Test update.

## Checklist


ontopy/patch.py
```diff
@@ -2,11 +2,17 @@
 
 # pylint: disable=protected-access
 import types
+from typing import Any, Generator, Tuple
```
jesper-friis, line 5: It is slightly more efficient to 

```
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from typing import Any, Generator, Tuple
```

and then change all annotations using these to strings. 

In Python 3.14 annotations are automatically converted to strings by the interpreter. But until then, we have to do it manually.

jesper-friis (approved):
Looks great. 

Very good with tests for `Metadata`. I see that there is no test for `Metadata.__repr__()`.

PR #864, opened 2025-06-15
Handle timeout errors in redirectioncheck

jesper-friis:
# Description
Instead of an unreadable traceback, report timeout as a standard failure.

**Question**: Should timeout be reported as success instead? 
If the request times out, it normally means that the redirection service is currently down, not that there necessary is something wrong with the redirections.

### DIFF
--- a/tools/redirectioncheck
+++ b/tools/redirectioncheck
@@ -52,12 +52,20 @@ def test_redirection(url, redirected_url, conf=None):
     Returns:
         ok: Whether the redirection was successful.
     """
-    verbose = conf.get("verbose")
-    r = requests.get(url, timeout=5)
-    ok, reason, got = r.ok, r.reason, None
+    if conf is None:
+        conf = {}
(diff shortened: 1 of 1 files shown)


tools/redirectioncheck
```diff
@@ -52,12 +52,20 @@ def test_redirection(url, redirected_url, conf=None):
...
-        ok, reason, got = False, "Invalid redirection", r.url
+    if conf is None:
+        conf = {}
+    verbose = int(conf["verbose"]) if "verbose" in conf else 1
+    timeout = float(conf["timeout"]) if "timeout" in conf else None
+
+    try:
+        r = requests.get(url, timeout=timeout)
+    except requests.Timeout:
+        ok, reason, got = False, "timeout", None
```
jesper-friis, line 63:
```suggestion
        ok, reason, got = True, "timeout", None
```

francescalb (approved):

PR #867, opened 2025-07-25
Added `--namespace` option to `ontoconvert`

jesper-friis:
# Description
Added a `--namespace` option to `ontoconvert`. This lets the user specify namespace prefixes in turtle output instead of the default `ns1`, `ns2`, ...

## Type of change
- [ ] Bug fix.
- [x] New feature.
- [ ] Documentation update.
- [ ] Test update.

## Checklist

### DIFF
--- a/ontopy/ontology.py
+++ b/ontopy/ontology.py
@@ -887,6 +887,7 @@ class Ontology(owlready2.Ontology):  # pylint: disable=too-many-public-methods
         overwrite=False,
         recursive=False,
         squash=False,
+        namespaces=None,
         write_catalog_file=False,
         append_catalog=False,
         catalog_file="catalog-v001.xml",
@@ -920,6 +921,9 @@ class Ontology(owlready2.Ontology):  # pylint: disable=too-many-public-methods
             If true, rdflib will be used to save the current ontology
             together with all its sub-ontologies into `filename`.
             It makes no sense to combine this with `recursive`.
+        namespaces: dict
+            Dict mapping prefixes to additional namespaces. Only used when
+            saving to turtle.
         write_catalog_file: bool
             Whether to also write a catalog file to disk.
         append_catalog: bool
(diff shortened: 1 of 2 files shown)


ontopy/ontology.py
```diff
@@ -936,6 +940,18 @@ def save(
         # pylint: disable=too-many-statements,too-many-branches
         # pylint: disable=too-many-locals,arguments-renamed,invalid-name
 
+        if namespaces is None:
+            namespaces = {}
+        default_namespaces = {
+            "": self.base_iri,
+            "locn": "http://www.w3.org/ns/locn#",
+            "swrl": "http://www.w3.org/2003/11/swrl#",
+            "bibo": "http://purl.org/ontology/bibo/",
```
francescalb, line 949: What is the reason for this choice? We have other, more often typically used namespaces.

ontopy/ontology.py
```diff
@@ -1083,9 +1095,13 @@ def save(
                 super().save(tmpfile, format="ntriples", **kwargs)
                 graph = rdflib.Graph()
                 graph.parse(tmpfile, format="ntriples")
-                graph.namespace_manager.bind(
-                    "", rdflib.Namespace(self.base_iri)
-                )
+
+                # Add common namespaces unknown to rdflib
```
francescalb, line 1099: Looks like we add all namespaces in the dict and override what is potentiallyin rdflib, no?

jesper-friis: Yes, that was the intention to allow the user to pontially override rdflib defaults.

ontopy/ontology.py
francescalb: OK, then add this information as a comment above.

ontopy/ontology.py
francescalb: OK, then update the comment to reflect this.

ontopy/ontology.py
```diff
@@ -1083,9 +1096,13 @@ def save(
...
```
jesper-friis, line 1100:
```suggestion
                # Add additional namespaces to the output graph
```

ontopy/ontology.py
jesper-friis: done

francescalb (approved):

ontopy/ontology.py
```diff
@@ -1032,14 +1050,9 @@ def save(
                 graph.add(triple)
 
             # Add common namespaces unknown to rdflib
```
jesper-friis, line 1052:
```suggestion
            # Add additional namespaces to the graph
```

PR #868, opened 2025-07-29
Deprecate support for python 3.8

francescalb:
# Description
Python3.8 reached end of life October 2024. It is assumed that all users of EMMOntoPy have passed to newer versions of python by now and that there is no need to spend time on issues pertaining only to py3.8.

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [ ] Test update.

## Checklist


requirements_dev.txt
jesper-friis, line 1: The `python_version>=3.9` part can be removed from all remaining lines.

requirements_docs.txt
```diff
@@ -1,8 +1,5 @@
 mike~=2.1
-mkdocs~=1.5; python_version<"3.8"
-mkdocs~=1.6; python_version>="3.8"
+mkdocs~=1.6; python_version>="3.9"
```
jesper-friis, line 2: We can skip the `python_version>=3.9` part to simplify the requirements file. Same for the other lines.

jesper-friis (commented):
Looks fine. But I think that we can simplify the requirement files even more by removing the `python_version>="3.9"` part.

requirements_dev.txt
francescalb: I kept it on purpose. I like to be specific and it shows clearly that on these packages there is probably an issue in lower versions of python if one wants to try that even tough we do no longer officially support it.

PR #909, opened 2025-12-11
Internal emmo

francescalb:
# Description
Made a copy of emmo in the local test directory in order to be less reliant on emmo, redirections and github.

Note that this does not significantly improve on the speed of the tests. For that we need a major clean-up.


jesper-friis (commented):
Looks nice. There might be a few other tests that must be updated when the new `emmopath` is used.

jesper-friis (approved):

PR #913, opened 2025-12-31
Remove owlready2 properties

jesper-friis:
# Description
Added option to ontoconvert to remove relations under the Owlready2 namespace.

### DIFF
--- a/emmopy/emmocheck.py
+++ b/emmopy/emmocheck.py
@@ -1022,10 +1022,9 @@ def main(
     if args.configfile:
         import yaml  # pylint: disable=import-outside-toplevel
 
-        with open(args.configfile, "rt") as handle:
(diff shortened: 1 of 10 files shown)


francescalb:
> > Nice, but should we use the remove_owlready2_properties also in the excel2onto parsers? Lines are 740 in ontopy/excelparser.
> 
> I don't think that it is necessary. Leaving the new `owlready2_properties` argument of Ontology.save() to False (default) will now skip Owlready2 properties when saving to turtle.
> 
> We probably don't even need to call remove_owlready2_properties() it in ontoconvert either...

Actually, this is not OK I jsut realised. In the excelparser there is logic that removes owlready2 names newly generated, but not those that come from imported ontologies. So saving should be done with owlready2_properties = True...

PR #946, opened 2026-03-24
Added a javascript for navigating to a section with a query parameter.

jesper-friis:
# Description
It is not possible to create redirection rules to URL fragments. 
To redirect an IRI of an ontological concept to its documentation we therefore have to use a query parameter.

This PR adds a javascript that, given the URI `page.html?section=my-heading` navigates to section "my-heading" in the similar way that `page.html#my-heading` would have done.

**WARNING**: untested code...

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [ ] Test update.

## Checklist


francescalb:
Fixed already in a different PR.

PR #949, opened 2026-04-10
Correct links

francescalb:
# Description
Current implementation is that the link is always created to the current page if the label is documented in the same page. 
This is wrong in cases where we have labeled something in the same way as something we refer to outside of the current ontology.

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [ ] Test update.

## Checklist

### DIFF
--- a/ontopy/ontodoc_rst.py
+++ b/ontopy/ontodoc_rst.py
@@ -474,7 +474,6 @@ class ModuleDocumentation:
                 if hasattr(entity, "get_annotations") or hasattr(
                     entity, "get_individual_annotations"
                 ):
-                    add_header("Annotations")
                     annotations = {  # pylint: disable=protected-access
                         a: a._get_values_for_class(  # pylint: disable=protected-access
                             entity
@@ -484,6 +483,8 @@ class ModuleDocumentation:
                             entity
                         )
                     }
+                    if len(annotations) > 0:
+                        add_header("Annotations")
 
                     long_annotations = [
                         "http://www.w3.org/2004/02/skos/core#example",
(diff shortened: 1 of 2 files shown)


ontopy/ontokit/setuptemplates/workflows/cd_release.yml
```diff
@@ -21,6 +20,20 @@ jobs:
...
+            echo "Missing .ontokit_conf.yml" >&2
+            exit 1
+          fi
+
+          value=$(sed -n -E "s/^ONTOLOGY_NAME:[[:space:]]*(.*)$/\\1/p" .ontokit_conf.yml | head -n 1)
+          if [ -z "$value" ]; then
+            echo "Missing required key 'ONTOLOGY_NAME' in .ontokit_conf.yml" >&2
+            exit 1
+          fi
+          echo "ONTOLOGY_NAME=${value}" >> "$GITHUB_ENV"
```
jesper-friis, line 35: This is a pattern that is repeated in all the yaml files. Would it make sense to make an `ontokit config` command such that this could be shortened to a single line

```bash
for def in $(ontokit config); do echo "$def" >> "$GITHUB_ENV"; done
``` 

that would set all the required config keys in one line and print an error message if any of the keys are missing?

ontopy/ontodoc_rst.py
```diff
@@ -298,7 +314,11 @@ def _html_links(full_iri, display_text):
             """Create the HTML code so that links lead to
             the correct fragment in the same document if possibe,
             otherwise link to the full IRI"""
-            fragment_iri = full_iri.split("#")[-1]
+            if full_iri not in page_entity_iris:
+                # Link to the full IRI if it's not documented on this page
+                return f"<a href='{full_iri}'>{display_text}</a>"
+
+            fragment_iri = getiriname(full_iri)
```
jesper-friis: `fragment_iri` may be a confusing variable name if the base of `full_iri` ends with a slash instead of a hash. `iriname` or just `name` may be less confusing.

jesper-friis (approved):
Impressive PR. Looks good, but it does quite much more than just correcting links.

ontopy/ontokit/setuptemplates/workflows/cd_release.yml
francescalb: I agree, and it has been commented before. But not part of this PR (wrong base branch, I am sorry). Should be fixed in a different PR.

francescalb:
> Impressive PR. Looks good, but it does quite much more than just correcting links.

Yes, ad you have already reviewed most of it before. I had put the wrong base_branch. Sorry!

PR #994, opened 2026-05-26
Emmocheckupdate

francescalb:
# Description

    Added test for numbers of rdfs:labels to be 1. This test is only run when enabled specifically.

    Added a test the check spelling with UpperCamelCase for classes only on prefLabels (not all labels as what we already have).
    This is run by default.

    Added a test the check spelling with UpperCamelCase for properties only on prefLabels (not all labels as what we already have).
    This is run by default.
    Note that this also widened what is allowed as last word after "is". I am not sure we should keep this part of the test though.

    The tests for CamelCase that were used previously are now skipped by default

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [ ] Test update.

## Checklist


emmopy/emmocheck.py
```diff
@@ -187,6 +216,71 @@ def test_object_property_label(self):
...
+        """
+        exceptions = set(
+            (
+                "0-manifold",  # not needed in 1.0.0-beta
+                "1-manifold",
+                "2-manifold",
+                "3-manifold",
+                "C++",
+                "3DPrinting",
+            )
```
jesper-friis, line 233: TODO: Remove these exceptions. They should in principle be added as local exceptions to EMMO, but can probably be dropped, since they are pre-EMMO 1.0.0.

jesper-friis:
It would be more elegant with just a single configurable `test_labels()`. The default configurations could be something like the following:

```python
labels_conf = {
      "ontology_labels": {"iri": "dcterms:title", "convention": None, "required": True, "unique": False},
      "class_labels": [
           {"iri": "skos:prefLabel", "convention": "UpperCamelCase", "required": True, "unique": True},
           {"iri": "skos:altLabel", "convention": "UpperCamelCase", "required": False, "unique": False},
           {"iri": "rdfs:label", "convention": None, "required": False, "unique": False},
    ],
    "property_labels": [
           {"iri": "skos:prefLabel", "convention": "lowerCamelCase", "required": True, "unique": True},
           {"iri": "skos:altLabel", "convention": "lowerCamelCase", "required": False, "unique": False},
           {"iri": "rdfs:label", "convention": None, "required": False, "unique": False},
    ],
    "individual_labels": [
           {"iri": "skos:prefLabel", "convention": "snake_case", "required": False, "unique": True},
           {"iri": "skos:altLabel", "convention": "snake_case", "required": False, "unique": False},
           {"iri": "rdfs:label", "convention": None, "required": False, "unique": False},
    ],
}
```

emmopy/emmocheck.py
```diff
@@ -187,6 +216,71 @@ def test_object_property_label(self):
...
+    def test_property_preflabel(self):
+        """Check that object property prefLabels are lowerCamelCase.
+
+        Allowed exceptions: "EMMORelation"
+
+        If they start with "has" or "is" they should be followed by a
+        upper case letter.
+
+        If they start with "is" they should also end with "Of", "With",
+        " After or "By".
```
jesper-friis:
```suggestion
        "After" or "By".
```

jesper-friis (approved):

PR #1000, opened 2026-05-29
Added test for uniqueness of labels

francescalb:
# Description
This check runs by default.
Is checks only for prefLabel as default.

One can add other labels if desired.

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [ ] Test update.

## Checklist

### DIFF
--- a/emmopy/emmocheck.py
+++ b/emmopy/emmocheck.py
@@ -90,9 +90,9 @@ class TestSyntacticEMMOConventions(TestEMMOConventions):
         namespace to `ignore_namespace` to avoid false positives.
 
         Configurations:
-            labels - annotation properties to validate [prefLabel].
-            exceptions - full names of entities to ignore.
-            ignore_namespace - namespaces to ignore..
+  - labels: annotation properties to validate. Default: `["prefLabel"]`.
+  - exceptions: full names of entities to ignore.
+  - ignore_namespace: namespaces to ignore..
         """
         testname = "test_unique_labels"
         exceptions = set()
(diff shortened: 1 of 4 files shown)


emmopy/emmocheck.py
```diff
@@ -82,6 +82,97 @@ def get_config(self, string, default=None):
...
+        """Check that configured labels are unique within each namespace.
+        This checks imported ontologies as well, but namespaces in the
+        `ignore_namespace` configuration are ignored. Thus, if you import
+        an ontology that you do not have control over, you can add its
+        namespace to `ignore_namespace` to avoid false positives.
+
+        Configurations:
+            labels - annotation properties to validate [prefLabel].
+            exceptions - full names of entities to ignore.
+            ignore_namespace - namespaces to ignore..
```
jesper-friis: For better rendering in documentation generators:

```suggestion
  - labels: annotation properties to validate. Default: `["prefLabel"]`.
  - exceptions: full names of entities to ignore.
  - ignore_namespace: namespaces to ignore..
```

emmopy/emmocheck.py
```diff
@@ -82,6 +82,97 @@ def get_config(self, string, default=None):
...
+                    continue
+                visited.add(entity)
+
+                r = repr(entity)
+                if r in exceptions or entity.iri.startswith(
+                    tuple(self.ignore_namespace)
+                ):
+                    continue
+
+                for lab in getattr(entity, label, []):
```
jesper-friis, line 134: Will this work with functional properties? 
If I remember correct will Owlready2 returns functional properties as a single value (and not a list).

jesper-friis (commented):
Added a comment

emmopy/emmocheck.py
francescalb: I have tested (Added it in test), seems to work.

jesper-friis (approved):

PR #1007, opened 2026-06-08
Nicer docs looks

francescalb:
# Description
New look

image

Vs current look:

image


jesper-friis:
There are a few conflicts that has to be resolved.

